PR #139, opened 2015-02-12
Make the unkown knowable

tjarratt:
This pull request fixes a typo that only the most pedantic person would bother fixing.

One small, annoying question I have is whether or not the error in `delta_tracker.go` should also be capitalized for consistency?

e.g.: `return fmt.Errorf("Unknown suite %s", suite.Path)`

It's not clear to me at first glance if that error is ever exposed back to the user.


onsi:
:) :+1:

Issue #141, opened 2015-02-16
Feature request: HTML reporter


abourget:
It would be very handy to have a simple HTML reporter.  Anyone wrote one somewhere ?

thanks!

onsi:
Not that I'm aware of, no.  Should be fairly easy though I think more useful would be a generic JSON output format that one could build tooling on top of for visualization.

Pushplaybang:
@abourget you should be able to use goconvey

Br3nda:
for those that land here and want to get a HTML report from your ginkgo tests, the magic command is:

```
go test ./... -json | go-test-report
```

you'll need to install https://github.com/vakenbolt/go-test-report

aasasi:
Hi @Br3nda ,
Have you had success using go-test-report  for GINKGO tests? Can `| go-test-report` be used with the `ginkgo` command?
I tried using this as you mentioned above, but it does not seem to recognize each spec. Any guidance on this is greatly appreciated.

Br3nda:
@[aasasi](https://github.com/aasasi) alas no. it sees each package as one test - it doesn't see each spec as it's own test.

however, this issue claims we can get json out put from ginkgo V2.
https://github.com/onsi/ginkgo/issues/639

onsi:
building out an html reporter for ginkgo would be super fun.  it hasn't been a priority for a while but with the new json format it wouldn't be too hard to implement something fairly feature rich.

it's a project i'd like to take on at some point.  i generally just use the console so this is a nice-to-have for me but i'd love to hear what kinds of usecases folks have in mind.  the most compelling one to me would be to be able to visualize and explore a failed CI test run in a browser after-the-fact.  but curious what others think.  i'd love to wrap my head around questions like enabling real-time browser updates vs a static page after the test runs (kinda how the pprof toolchain works).  is there value in running specs from the browser?  watching specs (ie `ginkgo watch`) from the browser?

onsi:
and one more clarification - the json format i'm reporting to is Ginkgo's own json format, not the go test json format.

aasasi:
Thanks @onsi ! As you mentioned, I would like to be able to visualize a summary of the test run and be able to drill down to the steps of each test for debugging . I have always loved the karate/cucumber test report , and I believe that something of that sort would be super helpful. With the current test logs, it is not super easy for folks who are not working on the tests to actually debug.

Mo0rBy:
I see this issue is closed but I would like to +1 it.
[...]

PR #143, opened 2015-02-19
Add a Json reporter

Etienne42:
Implements Json reporter
Any comments welcome

Tests: Ran 52 of 52 Specs in 0.480 seconds

Reference: https://github.com/onsi/ginkgo/issues/141
cc: @abourget

### DIFF
--- a/reporters/json_reporter.go
+++ b/reporters/json_reporter.go
@@ -1,0 +1,126 @@
+package reporters
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/onsi/ginkgo/config"
+	"github.com/onsi/ginkgo/types"
+)
+
+type JsonReporter struct {
+	suite         JsonTestSuite
+	filename      string
+	testSuiteName string
+}
+
+type JsonTestSuite struct {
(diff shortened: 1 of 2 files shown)


onsi:
Awesome!  I'll try to take a look soon.  I'm thinking of adding "first class" support for this whereby someone could run `--json` and see JSON output instead of the default test runner output.

That way `ginkgo -p --json` will also allow us to emit JSON in parallel (the story around how parallel runs work with custom reporters is woefully incomplete, sadly).

Thanks so much!

onsi:
Finally taking a look now.

I think it would be more flexible if the reporter were passed an io.Writer instead of a filename.  I'll make that change and then merge it in.  Thanks!

onsi:
Hmm.. so looking through it looks like this reporter collects test state and only emits output - as a single JSON document - at the very end.

There's a longer term vision that I have in mind in which tooling could be built on top of Ginkgo to handle streaming output.  I was thinking that a JSON reporter would emit little snippets of JSON as the tests ran.

onsi:
Am going through and cleaning up old issues/PRs.  Closing this one.  Please reopen it if you would like.

PR #152, opened 2015-04-23
Implemented the ability to specify a command after a suite runs

johnnadratowski:
This is the second of my PRs I'd like to put forth for consideration.  This PR allows the user to run an arbitrary command after a suite runs.

I added some string replacement to pass input to the command. 

Unfortunately, I don't have access to an OSX box to test this on.


ginkgo/notifications.go
```diff
@@ -59,3 +61,27 @@ func (n *Notifier) SendNotification(title string, subtitle string) {
 		exec.Command("terminal-notifier", args...).Run()
 	}
 }
+
+func (n *Notifier) RunCommand(suite testsuite.TestSuite, suitePassed bool) {
```
johnnadratowski: Not sure if this is the right place for this, but it was the only place I saw with access to the command line args from the suite completion.

johnnadratowski:
@onsi - Just mentioning you for a notification, not sure if you saw these PRs or not

onsi:
hey @johnnadratowski 

first off - thanks for the PR.  I'd like to dig in a bit into why you want this.  What problems are you trying to solve, etc.  It's possible that your usecase might be satisfied by a custom reporter (though I might be wrong).

Not to say that I won't pull it in or anything -- just wanting to understand first.  The `ginkgo` CLI is becoming something of a kitchen sink (which isn't *necessarily* bad!).

Also - if you could add an [integration test](https://github.com/onsi/ginkgo/tree/master/integration) that would be great.  I don't have much test coverage in terms of CLI unit tests (ironically) but just about everything is covered in an integration tests

johnnadratowski:
@onsi - Thanks for the reply.  My first PR is pretty self-explanatory - I just wanted to get a notification on Linux.  In terms of this one, I was struggling with how to pass the notifications out of my vagrant machine to my local.  My idea was that I could ssh back to the host box and run the notification command from there, but I would need some custom command to run after the suite passed.

It was pretty late when I finished this, so I wasn't able to actually try it yet, but I think it should work.

johnnadratowski:
@onsi - Hey, had a sick day so I was able to finally add some tests to this.  I also added the output from the suite command properly reporting success/failure.  I merged in the latest master and there were some minor conflicts I had to resolve, but all the tests indicate its working properly.

onsi:
Hey @johnnadratowski  - sorry again for the delay.  This is starting to grown on me.  Could you do a couple of things before I merge it in?

1. `--command` is a bit generic.  How about `--afterSuiteHook` or `--runAfterSuite` instead?  (Also, apologies for the camel-case but it's consistent with the rest of the CLI).
2. Do you mind updating the gh-pages branch to document this?

Thanks!

johnnadratowski:
@onsi - Cool! I'm really happy you like it.  We have a big deadline this week at work, so it might take me a few days, but I should be able to get to this by EOW, at the latest.

Thank you for merging my other PRs!  Especially the coverpkg one, helps me out a bunch with some other projects.

onsi:
Sorry for the delay - thanks for the PR!

PR #153, opened 2015-04-24
Use two dashes for test.timeout argument.

ixdy:
Alternate flag libraries such as github.com/spf13/pflag add support for
shorthand flags. This causes `-test.timeout` to be interpreted as the
shorthand flag `t` rather than the longhand flag `test.timeout`.

Using two dashes avoids this issue.


onsi:
thanks!

Issue #156, opened 2015-05-04
Ginkgo does not work with go <=1.3


wwwdata:
Hi,

we are using ginkgo for a long time now and it always worked fine. But it seems that the newest version does not work correctly with go 1.3 and 1.2. It works all fine with 1.4

The tests are working when executed with `go test`. But it is not possible to run them with the ginkgo command line tool. But we'd like to do that for additional options for our CI builds.

Output:
```bash
root@19a97b3d9245:/gopath/src/github.com/manyminds/api2go# go version
go version go1.3 linux/amd64
root@19a97b3d9245:/gopath/src/github.com/manyminds/api2go# ginkgo
Failed to run test suite!
	fork/exec /tmp/ginkgo721761102/api2go.test: no such file or directory
Ginkgo ran 1 suite in 942.915467ms
Test Suite Failed
root@19a97b3d9245:/gopath/src/github.com/manyminds/api2go# go test
Running Suite: Api2go Suite
===========================
Random Seed: 1430747014
Will run 56 of 56 specs

••••••••••••••••••••••••••••••••••••••••••••••••••••••••
Ran 56 of 56 Specs in 0.020 seconds
SUCCESS! -- 56 Passed | 0 Failed | 0 Pending | 0 Skipped PASS
ok  	github.com/manyminds/api2go	0.110s
```

Here is a travis build https://travis-ci.org/manyminds/api2go/builds/61144609

wwwdata:
Hm, just saw in the Readme go version 1.4+

Does that mean older versions are not supported anymore?

onsi:
Hey @wwwdata - yes I introduced a change that requires go 1.4.  I'm using go test -c -o which is only available in go 1.4.

If this is very painful for you/others I can fairly easily add a fallback that will work with older versions.  Was trying to avoid that if possible.

wwwdata:
alright, thanks for your reply, i will close the ticket

fgrosse:
I would like to have the fallback if it is easily possible. I love ginko but not all my systems do have go 1.4 yet and I would like to support go 1.3 a little longer. Until then I will now stick with an older version of ginkgo

mwf:
Hi!
The same thing for me - I'd like to have the fallback. Debian stable repos provide go 1.3, that's why it's still widely used.

onsi:
It's not hard to do but I'm pretty busy these days :/

Is anyone up for a PR?  I'd be happy to point you in the right direction.

[...]

PR #166, opened 2015-05-31
go 1.3 compatibility

fgrosse:
As discussed in #156 this fixes ginkgo when running on go 1.3 where the `-o` flag of test does not work.

If you run `go test -c -o foo.test` on an older go version it will not cause an error but just silently ignore the flag. I resolved the issue by checking if the compilation was successful but the compiled output did not exist on the requested target destination. If this is the case we can assume that `o` did not work and move the file by hand.

The code should bve covered by travis since I added older go versions in the build matrix.
Before I could start I had to fix another issue that made my tests fail (6412c76).
The problem there was that the actual test output was not in the expected format and sometimes failed.

If you want I can split up this PR into two.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: go
 go:
-  - 1.2
   - 1.3
   - 1.4
   - tip
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,9 @@
 language: go
 go:
+  - 1.2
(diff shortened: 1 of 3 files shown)


fgrosse:
Note: the tests still fail on go 1.2 so I had to remove that from the travis build matrix

onsi:
Brilliant!  Thanks @fgrosse

PR #167, opened 2015-06-09
`panic(nil)` in a test results in it immediately passing

sclevine:
This behavior was identical to that of the built-in `testing` package, but it changed over a year ago: https://github.com/golang/go/issues/6546

Relevant sections:
 - [internal/leafnodes/runner.go#L97](https://github.com/onsi/ginkgo/blob/master/internal/leafnodes/runner.go#L97)
 - [internal/leafnodes/runner.go#L72](https://github.com/onsi/ginkgo/blob/master/internal/leafnodes/runner.go#L72)
 - [ginkgo_dsl.go#L248](https://github.com/onsi/ginkgo/blob/master/ginkgo_dsl.go#L248)

This makes it easy to unknowingly skip assertions, so it seems pretty dangerous. That said, fixing it may break some people's test suites.

(CC @rosenhouse)

### DIFF
--- a/internal/leafnodes/shared_runner_test.go
+++ b/internal/leafnodes/shared_runner_test.go
@@ -96,6 +96,24 @@ func SynchronousSharedRunnerBehaviors(build func(body interface{}, timeout time.
 				Ω(failure.ForwardedPanic).Should(Equal("ack!"))
 			})
 		})
+
+		Context("when a panic occurs with a nil value", func() {
+			BeforeEach(func() {
+				outcome, failure = build(func() {
+					didRun = true
+					innerCodeLocation = codelocation.New(0)
+					panic(nil)
+				}, 0, failer, componentCodeLocation).Run()
+			})
+
+			It("should return the nil-valued panic", func() {
+				Ω(didRun).Should(BeTrue())
+
+				Ω(outcome).Should(Equal(types.SpecStatePanicked))
+				Ω(failure.ForwardedPanic).Should(Equal("<nil>"))
+			})
+		})
+
 	})
 }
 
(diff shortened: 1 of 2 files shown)


sclevine:
Done. Let me know what you think.

A few points:

1. Some people may be using `panic(nil)` intentionally. It could be considered a convenient (although highly confusing) way of short-circuiting the rest of a test.

2. I don't think it's possible to fix `GinkgoRecover()` without changing its interface.

internal/leafnodes/shared_runner_test.go
```diff
@@ -96,6 +96,24 @@ func SynchronousSharedRunnerBehaviors(build func(body interface{}, timeout time.
...
+					innerCodeLocation = codelocation.New(0)
+					panic(nil)
+				}, 0, failer, componentCodeLocation).Run()
+			})
+
+			It("should return the nil-valued panic", func() {
+				Ω(didRun).Should(BeTrue())
+
+				Ω(outcome).Should(Equal(types.SpecStatePanicked))
+				Ω(failure.ForwardedPanic).Should(Equal("<nil>"))
```
rosenhouse: is this preferable to a more explicit, hardcoded error like `test executed panic(nil) or runtime.Goexit` [in the standard library](https://github.com/golang/go/blob/c9778ec302b2e0e0d6027e1e0fca892e428d9657/src/testing/testing.go#L444)?

rosenhouse: Ok, it's getting wrapped with the "Test Panicked" message in `failer.go`.  That looks good.

sclevine: I went back and forth on whether this message should be more verbose. On one hand, `<nil>` is consistent with how Ginkgo normally behaves when you panic (and as you mentioned, it indicates that test panicked along with the `nil` value). On the other hand, you'll get this same message when you call `runtime.Goexit()`, and that could be confusing.

onsi:
yep not much to be done with `GinkgoRecover()` - it's typically run in a spawned Goroutine and it's impossible to recover a panic in one Goroutine from another so there's no way to play the `finished` trick without changing the external interface.

PR #171, opened 2015-07-31
This fix resolves an appeareance issue when spec is skipped. 

svett:
The skipped spec and the reason message should be in cyan color instead of red color.


onsi:
Thanks for this. Super minor nit: could we make IsFailureState a method on SpecState instead of a package method?

svett:
Good suggestion. Please check new commit that address this.

onsi:
Thanks for braving the terribleness that is stenographer.go

svett:
Many thanks!

PR #182, opened 2015-09-02
small typo fix for table documentation

jbayer:


robdimsdale:
@jbayer I think the docs are correct as they are. They refer to [dot importing](https://golang.org/ref/spec#Import_declarations) which makes interacting the with functions slightly more readable and fluent i.e.:

```
import (
    . "github.com/onsi/ginkgo/extensions/table"
)
```

allows the exported functions in the `table` package to be referenced with prepending `table.SomeFunction`

jbayer:
ok, i'm a total n00b.

PR #190, opened 2015-10-07
TestRunner copies compiled file if moving it fails.

### LINKED ISSUE #189: Copy compiled file if move fails
When running on golang < 1.4, there is no `-o` option so ginkgo attempts to compile the file and move it (see [test_runner.go](https://github.com/onsi/ginkgo/blob/462326b1628e124b23f42e87a8f2750e3c4e2d24/ginkgo/testrunner/test_runner.go#L108))

This fails if `/tmp` is on a different partition. It should be possible to copy the file across partitions instead.

I appreciate this is an edge-case as not many people will be running tests on golang 1.2 and 1.3, but I'm supporting a library on all versions of golang, and currently ginkgo is failing.

How do you feel about adding a `io.Copy` if `os.Rename` fails? I'm willing to submit a PR for this, but I'm a little concerned about how to drive this out with a test, as forcing `os.Rename` to fail could be tricky.

robdimsdale:
- This fixes an edge case when running on go < 1.4 and moving the file fails,
  perhaps because the source and destination are on different partitions.
- Fixes #189

### DIFF
--- a/ginkgo/testrunner/test_runner.go
+++ b/ginkgo/testrunner/test_runner.go
@@ -110,7 +110,9 @@ func (t *TestRunner) CompileTo(path string) error {
 				// We cannot move the file, perhaps because the source and destination
 				// are on different partitions. We can copy the file, however.
 				err = copyFile(compiledFile, path)
-				return fmt.Errorf("Failed to copy compiled file: %s", err)
+				if err != nil {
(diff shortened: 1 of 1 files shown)


onsi:
woot?

vito:
Don't you also need the copied file to be executable?

robdimsdale:
@vito yes I think so. I'm looking into making that change now

PR #191, opened 2015-10-08
Ensure copied compiled file is executable

robdimsdale:
ginkgo requires the `.test` file to be executable - when we copied the contents before we didn't also make the copied file executable.

This should close out #190.


robdimsdale:
FYI @vito is reviewing this - on his suggestion I'm going to preserve the existing file's permissions - more commits to come.

robdimsdale:
@vito says it looks good so I'm going to merge this once the tests are green.

onsi:
Sweet!  Thanks both.

PR #197, opened 2015-11-26
Ginkgo unfocus should work for DescribeTable and Entry

ebabani:
Currently they're being ignored.

Also added a test for unfocusing Measure.


onsi:
Woot!  Thanks @ebabani!

PR #203, opened 2015-12-11
Don't print slowSpecThreshold default twice

krishicks:
ginkgo help |& grep -A1 slowSpecThreshold

Before:
  -slowSpecThreshold float
    .. by the default reporter (default: 5 seconds). (default 5)

After:
  -slowSpecThreshold float
    .. by the default reporter. (default 5)


onsi:
Thanks!

PR #204, opened 2015-12-15
Support GO15VENDOREXPERIMENT by ignoring vendored test suites.

xeger:


xeger:
Grr! It appears that `os.LookupEnv` is a newer part of stdlib.

If you agree with this approach, let me know and I'll reimplement using os.Getenv (ugh).

arangamani:
:+1: 
This way I can get rid of the following from my bash aliases:
```
alias ginkgo='ginkgo -skipPackage vendor'
```

onsi:
First off: thank you.
Second: double thank you for maintaining the randomn Clue-themed test setup.
Thirdly: I'd like to maintain support for 1.4 so we'll need to use Getenv instead.  Sorry...

Thanks!

xeger:
Done; supporting 1.4 requires us to be a bit more pedantic about the value of `GO15VENDOREXPERIMENT` but with a bit of luck the experiment will succeed and I'l be able to get rid of the check for language versions >= 1.6.

xeger:
Whoah! Evidently, 1.3 lacks `os.Unsetenv` .. presumably you'd like to support 1.3 as well, and this is a test-only thing. Will fix in a bit.

xeger:
Third time's the charm. (I force-pushed my second commit with a fix for the Go 1.3 issue.) I'll revisit this once Go 1.6 is released. Thank you @onsi for making testing in Golang not merely tractable, but pleasant! I look forward to contributing more in the future.

onsi:
looks good - will merge it in though it might be worth revisting this soon now that 1.6 is in beta?

Thanks for the help!  Glad you're enjoying Ginkgo.

PR #207, opened 2016-01-08
Generate should use the same package as the target file

dprotaso:
Otherwise it leads to the practice of exporting more types in your package than necessary

### DIFF
--- a/ginkgo/generate_command.go
+++ b/ginkgo/generate_command.go
@@ -30,7 +30,7 @@ func BuildGenerateCommand() *Command {
 	}
 }
 
-var specText = `package {{.Package}}_test
+var specText = `package {{.Package}}
 
 import (
 	. "{{.PackageImportPath}}"


jasonkeene:
+:100:  :shipit:

robdimsdale:
Doesn't this mean that generated tests will have access the unexported types in the code under test?

krishicks:
-1

jim-slattery-rs:
I'd prefer using the `<package>_test` by default.
That way your tests are free to import things that might otherwise cause an import cycle if they were in the same `<package>`.

If there are private things in your package to test, there's nothing stopping you from creating a <package>_private_test.go file to test those private members, when needed.

krishicks:
It might be good for @dprotaso to show an example of why this would be beneficial.

You should only be testing exported structs and exported functions on those structs. In what scenario do you need access to an unexported struct when using exported functions?

tjarratt:
-1

I strongly believe that Ginkgo should encourage users to only test their public interface. This pattern does encourage you to expose more things, but it also forces you to think about what you're exposing and why, and encourages you to write smaller packages.

ryanmoran:
This encourages poor testing practices. -1

Amit-PivotalLabs:
-1. I want confidence that my tests prove that the exported interface of my package works, and to that end I want confidence that my tests can't access the unexported interface.

onsi:
@dprotaso the use of the `_test` package as the default was intentional precisely for the reasons enumerated on this thread.  Given how easy it is to remove the `_test` suffix I think this is a reasonable opinionated default that folks can opt out of easily if they need to.

I don't want the default to change.  I'd be happy to accept a PR that adds a flag to both generate and bootstrap to allow users to generate files without the test suffix.

dprotaso:
@krishicks 

I've been dealing with code that involves timers and to make tests deterministic and run faster I've abstract the timer so it can be mocked. I don't want to leak this abstraction to the consumers of my package hence my unit tests would be calling internal constructors.

I don't want the consumers of my package to use https://github.com/pivotal-golang/clock

@onsi Yeah I agree with everyone's points - but occasionally things do end up being exported

ie. The following files export an interface and the implementing struct
https://github.com/cloudfoundry/cli/blob/master/cf/api/app_files/app_files.go#L14
https://github.com/cloudfoundry/cli/blob/master/cf/api/space_quotas/space_quotas.go#L28

In the extreme cases it's probably best to use internal packages https://golang.org/doc/go1.4#internalpackages

krishicks:
@dprotaso The cli code is probably the farthest thing from an exemplary Golang codebase. That the struct is exported is a bad thing, and not necessary at all.

Issue #208, opened 2016-01-16
Not building on arm64


luxas:
`syscall.Dup2` doesn't exist when building on `linux_arm64`...
Kubernetes uses this code and now when I'm trying to get it working on `arm64`, this fails.

Could you add an alias or something? Because `syscall.Dup3(oldfd, newfd, 0)` works on arm64.
I will return with more information soon and maybe a PR. 
Cross-ref: golang/go#10235

luxas:
Fixed with: #209

PR #209, opened 2016-01-16
Add linux_arm64 support

luxas:
Ref: #208

This patch adds support for `linux_arm64` support. 
I'm working on Kubernetes support for `arm64` and Kubernetes uses this project, so if we could merge this it would be great!

The difference between `Dup2` and `Dup3` is here: http://linux.die.net/man/2/dup3
Quick compilation test:
```
$ GOARCH=arm64 GOOS=darwin go build github.com/onsi/ginkgo
$ GOARCH=arm GOOS=darwin go build github.com/onsi/ginkgo
$ GOARCH=amd64 GOOS=darwin go build github.com/onsi/ginkgo
$ GOARCH=amd64 GOOS=netbsd go build github.com/onsi/ginkgo
$ GOARCH=amd64 GOOS=linux go build github.com/onsi/ginkgo
$ GOARCH=arm64 GOOS=linux go build github.com/onsi/ginkgo
$ GOARCH=arm GOOS=linux go build github.com/onsi/ginkgo
```
Build completed successfully on all tests. Earlier `GOARCH=arm64 GOOS=linux` failed.
But I haven't tested it on `arm64` (as I don't have a board yet), but I think it should be OK
Please say what you think!

### DIFF
--- a/internal/remote/syscall_dup_unix.go
+++ b/internal/remote/syscall_dup_unix.go
@@ -1,0 +1,10 @@
+// +build !linux !arm64
+// +build !windows
+
+package remote
+
+import "syscall"
+
+func syscallDup(oldfd int, newfd int) (err error) {
+	return syscall.Dup2(oldfd, newfd)
+}
--- a/internal/remote/output_interceptor_unix.go
+++ b/internal/remote/output_interceptor_unix.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"io/ioutil"
 	"os"
-	"syscall"
 )
 
 func NewOutputInterceptor() OutputInterceptor {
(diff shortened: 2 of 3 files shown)


onsi:
Sorry for the delay -- will this break on windows?  (see my comment inline on the PR)

internal/remote/syscall_dup_unix.go
```diff
@@ -0,0 +1,9 @@
+// +build !linux !arm64
```
onsi: Should this be
// +build !linux !arm64 !windows

?

luxas: Yes, of course! It's a bit hard to test on all platforms :)
Now the formula is:
`((NOT linux) OR (NOT arm64)) AND (NOT windows)`
and that's equalient to 
`(NOT linux,arm64) AND (NOT windows)`

Now it works on windows too...

luxas:
Thanks for the pointer. Updated

onsi:
Great - thanks!

PR #212, opened 2016-01-27
We no longer pass untyped nils to the reflect::Value.Call

dprotaso:
This would cause panics

Signed-off-by: Topher Bullock <cbullock@pivotal.io>

### DIFF
--- a/extensions/table/table_entry.go
+++ b/extensions/table/table_entry.go
@@ -23,8 +23,17 @@ func (t TableEntry) generateIt(itBody reflect.Value) {
 	}
 
 	values := []reflect.Value{}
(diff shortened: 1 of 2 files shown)


dprotaso:
ie. this no longer panics
```
package broken_test

import (
    . "github.com/onsi/ginkgo"
    . "github.com/onsi/ginkgo/extensions/table"
)

var _ = Describe("table with nils", func() {
    DescribeTable("no panics!",
        func(val *bool) {
        },
        Entry("broken", nil),
    )
})
```

onsi:
Nice - thanks y'all!

PR #216, opened 2016-01-29
Specify block additions to DSL

sha1sum:
In certain cases, `It` does not fit in a natural sentence flow with the surrounding blocks. In these cases, `Specify` would be more natural to read. An example of such blocks would be:

```go
Describe("The foobar service", func() {
  Context("when calling Foo()", func() {
    Context("when no ID is provided", func() {
      Specify("an ErrNoID error is returned", func() {
      })
    })
  })
})
```

The new functions `Specify`, `FSpecify`, `PSpecify`, and `XSpecify` simply pass their arguments to the corresponding `It`, `FIt`, `PIt`, or 'XIt` functions.

I have taken a look around ginkgo's internal test specs and I'm not sure where unit tests would go for these additions, but if someone would like to point me in the right direction I'd be happy to add those and update the pull request.

### DIFF
--- a/ginkgo_dsl.go
+++ b/ginkgo_dsl.go
@@ -347,6 +347,28 @@ func XIt(text string, _ ...interface{}) bool {
 	return true
 }
 
+//Specify blocks are aliases for It blocks and allow for more natural wording in situations
+//which "It" does not fit into a natural sentence flow. All the same protocols apply for Specify blocks
+//which apply to It blocks.
+func Specify(text string, body interface{}, timeout ...float64) bool {
+	return It(text, body, timeout...)
+}
+
+//You can focus individual Specifys using FSpecify
+func FSpecify(text string, body interface{}, timeout ...float64) bool {
+	return FIt(text, body, timeout...)
+}
+
+//You can mark Specifys as pending using PSpecify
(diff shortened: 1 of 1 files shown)


robdimsdale:
@sha1sum looks fine, though I personally find myself rewording the text in an `It` block such that the action is first, e.g. in the example you provide I would reword from:

```
Context("when no ID is provided", func() {
    Specify("an ErrNoID error is returned", func() {
    })
})
```

to

```
Context("when no ID is provided", func() {
    It("returns an ErrNoID error", func() {
    })
})
```

sha1sum:
@robdimsdale to me, if the strings of the blocks are read as a sentence, the specify example  provides a slightly better sentence structure. It's not a huge difference, and I'm sure it's a matter of preference, but slightly easier to read in my opinion.

robdimsdale:
Totally reasonable. If @onsi has no objections, I'd like to see updates to the documentation and then it could be merged

onsi:
SGTM but please do update the docs on the gh-pages branch.

sha1sum:
@onsi Added [PR #217](https://github.com/onsi/ginkgo/pull/217) for documentation.

PR #217, opened 2016-01-29
Documentation Updates for Addition of `Specify` in DSL

sha1sum:
This is a complement to [PR #216](https://github.com/onsi/ginkgo/pull/216) to add documentation for the addition of the `Specify` alias to `It` to the DSL.

### DIFF
--- a/index.md
+++ b/index.md
@@ -142,8 +142,8 @@ Let's break this down:
 
 - Ginkgo makes extensive use of closures to allow you to build descriptive test suites.
 - You should make use of `Describe` and `Context` containers to expressively organize the behavior of your code.
-- You can use `BeforeEach` to set up state for your specs.  You use `It` to specify a single spec.  
-- In order to share state between a `BeforeEach` and an `It` you use closure variables, typically defined at the top of the most relevant `Describe` or `Context` container.
+- You can use `BeforeEach` to set up state for your specs.  You use `It`/`Specify` to specify a single spec. `It` and `Specify` have identical behavior. 
+- In order to share state between a `BeforeEach` and an `It`/`Specify` you use closure variables, typically defined at the top of the most relevant `Describe` or `Context` container.
 - We use Gomega's `Expect` syntax to make expectations on the `CategoryByLength()` method.
 
 Assuming a `Book` model with this behavior, running the tests will yield:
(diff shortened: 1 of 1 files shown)


onsi:
Thanks for this @sha1sum - I can see that you put some time into it.  I see you've peppered `Specify` throughout the docs and am wondering if that is clearer than having a separate section that simply describes Specify as an alias...  @robdimsdale any thoughts?

robdimsdale:
Yes, thanks @sha1sum for the docs update. I think I agree with @onsi that it might be clearer to specify the alias in a single place, rather than multiple places throughout the docs. Perhaps something similar to how `By()` is documented: https://onsi.github.io/ginkgo/#documenting-complex-its-by

Or maybe it doesn't even need to be in its own section, it could just be mentioned once under the `It()` heading: https://onsi.github.io/ginkgo/#individual-specs-it

sha1sum:
@onsi and @robdimsdale - I've updated PR so that the `Specify` documentation is condensed to one section with a subheading under the `It` block section. Let me know if you need any further changes :)

robdimsdale:
@sha1sum this is great, and thanks for cleaning up all the whitespace too!

Would it be possible for you to add an example as to when this alias might be useful? The original snippet you posted in https://github.com/onsi/ginkgo/pull/216 would work well:

```
Describe("The foobar service", func() {
  Context("when calling Foo()", func() {
    Context("when no ID is provided", func() {
      Specify("an ErrNoID error is returned", func() {
      })
    })
  })
})
```

Once we see that I'd consider these docs good to go, and unless @onsi has any objections we can merge them.

sha1sum:
@robdimsdale I've added the `Specify` example, and I've also changed the multi-line go code blocks to use "\`\`\`go" and "\`\`\`" so that they're properly formatted on GitHub. If I've overstepped on that one, I can set it back, but the syntax should work with pretty much any markdown processor and has the added benefit of syntax highlighting on GitHub-flavored markdown parsers.

robdimsdale:
@sha1sum it looks like your last commit touched about every line in that file so it's hard to see exactly what's changed in the GitHub UI. I'm taking a look locally, and I like the example you gave for `Specify` - I'd be happy to see that merged.

For the code blocks, I'm not sure how they render in github pages, and I'm not sure how to test that before merging - so I'd probably prefer to see that in a separate commit so we can isolate it. Thoughts, @onsi?

onsi:
Sadly - unless things have changed recently (and thy may well have!) I believe the triple-backtick code blocks don't work on gh-pages?  Though that may be my fault in how I've configured things?

sha1sum:
Sorry about that, @onsi and @robdimsdale -- I've just reverted the file back to the `gh-pages` revision and then re-added the `Specify` subsection with example use case.

robdimsdale:
I'll take a look shortly, thanks.

Also I wonder if the markdown rendering might work better now that github is using jekyll 3:

https://github.com/blog/2100-github-pages-now-faster-and-simpler-with-jekyll-3-0

robdimsdale:
@sha1sum thanks, and apologies for the back-and-forth. This example looks good to me, happy to merge if @onsi has no objections

separately, I'm curious about the docs so I opened an issue to investigate the possibility of syntax highlighting: https://github.com/onsi/ginkgo/issues/220

onsi:
Lgtm!

robdimsdale:
And it's live! Thanks @sha1sum

PR #225, opened 2016-02-26
Fix headlines.

gabrielf:


gabrielf:
Currently many of the headlines and links to headlines such as: http://onsi.github.io/ginkgo/#avoiding-dot-imports don't work because of missing spaces.

onsi:
Thanks!

PR #227, opened 2016-02-27
Add ability to dump GinkgoWriter on the fly with SIGUSR1

benmoss:
Hey @onsi! 

This is a feature addition to make it so users can signal the Gingko process with SIGUSR1 to flush all the logs the GinkgoWriter has buffered so far. This is useful in the case where you have a test taking longer than expected but don't want to interrupt it, just want to inspect it.

The one open question I have is whether it should flip the bit to switch the GinkgoWriter into verbose/stream mode. Right now it's just flushing and leaving that alone, but I could see the other being useful and maybe even expected.

Thoughts?

Also did a bunch of copy/paste to cobble this together, so feel free to complain wildly about that and maybe I can clean things up.

### DIFF
--- a/integration/verbose_and_succinct_test.go
+++ b/integration/verbose_and_succinct_test.go
@@ -1,11 +1,14 @@
 package integration_test
 
 import (
+	"os/exec"
 	"regexp"
 	"runtime"
+	"syscall"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gbytes"
 	"github.com/onsi/gomega/gexec"
 )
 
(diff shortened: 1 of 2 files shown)


benmoss:
Test failure seems to just be flakiness

onsi:
hey @benmoss this is very cool.  I'd love for SIGUSR1 to flip the `--v` bit to on.

A few questions:
- SIGUSR1 isn't a thing on windows is it?  probably ok.
- this probably doesn't work when running in parallel, right?  I think that would be important as most folks experience stuck tests when running in parallel

benmoss:
Nope, there isn't a SIGUSR1 on Windows. Someone could probably figure out a way to do something similar for Windows, but I suspect it'd be a lot harder.

I'll look into how this works with --parallel.

svett:
@benmoss you can move the POSIX implementation related to SIGUSR1 in a separate file. Then you can use the [Golang build tag](http://blog.ralch.com/tutorial/golang-conditional-compilation/)s for POSIX operation systems. You can keep a noop file for Windows.

benmoss:
Yep, I'm gonna close this though since I don't have a lot of time to dig into the parallel spec runner implementation.

PR #228, opened 2016-02-29
Regex based file filtering : Feature allowing us to skip based on cod…

jayunit100:
@onsi Ok here we go ! 

Can you help me write the unit test in a follow on patch?    I hacked around a little with it, but it seemed a little tricky how the actual tests were working for spec.go, and ive got a few other priorities on hand.

This will be really useful in upstream kubernetes and openshift for us.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -23,15 +23,16 @@ import (
 const VERSION = "1.2.0"
 
 type GinkgoConfigType struct {
-	RandomSeed        int64
-	RandomizeAllSpecs bool
-	FocusString       string
-	SkipString        string
-	SkipMeasurements  bool
-	FailOnPending     bool
-	FailFast          bool
-	EmitSpecProgress  bool
-	DryRun            bool
+	RandomSeed         int64
+	RandomizeAllSpecs  bool
+	RegexScansFilePath bool
+	FocusString        string
+	SkipString         string
+	SkipMeasurements   bool
+	FailOnPending      bool
+	FailFast           bool
+	EmitSpecProgress   bool
+	DryRun             bool
 
 	ParallelNode  int
 	ParallelTotal int
@@ -66,9 +67,14 @@ func Flags(flagSet *flag.FlagSet, prefix string, includeParallelFlags bool) {
 	flagSet.BoolVar(&(GinkgoConfig.SkipMeasurements), prefix+"skipMeasurements", false, "If set, ginkgo will skip any measurement specs.")
 	flagSet.BoolVar(&(GinkgoConfig.FailOnPending), prefix+"failOnPending", false, "If set, ginkgo will mark the test suite as failed if any specs are pending.")
 	flagSet.BoolVar(&(GinkgoConfig.FailFast), prefix+"failFast", false, "If set, ginkgo will stop running a test suite after a failure occurs.")
+
 	flagSet.BoolVar(&(GinkgoConfig.DryRun), prefix+"dryRun", false, "If set, ginkgo will walk the test hierarchy without actually running anything.  Best paired with -v.")
+
 	flagSet.StringVar(&(GinkgoConfig.FocusString), prefix+"focus", "", "If set, ginkgo will only run specs that match this regular expression.")
 	flagSet.StringVar(&(GinkgoConfig.SkipString), prefix+"skip", "", "If set, ginkgo will only run specs that do not match this regular expression.")
+
+	flagSet.BoolVar(&(GinkgoConfig.RegexScansFilePath), prefix+"regexScansFilePath", false, "If set, ginkgo regex matching also will look at the file path (code location).")
+
 	flagSet.BoolVar(&(GinkgoConfig.EmitSpecProgress), prefix+"progress", false, "If set, ginkgo will emit progress information as each spec runs to the GinkgoWriter.")
 
 	if includeParallelFlags {
@@ -140,6 +146,10 @@ func BuildFlagArgs(prefix string, ginkgo GinkgoConfigType, reporter DefaultRepor
 
 	if ginkgo.SyncHost != "" {
 		result = append(result, fmt.Sprintf("--%sparallel.synchost=%s", prefix, ginkgo.SyncHost))
+	}
+
+	if ginkgo.RegexScansFilePath {
+		result = append(result, fmt.Sprintf("--%sregexScansFilePath", prefix))
 	}
 
 	if reporter.NoColor {
--- a/integration/flags_test.go
+++ b/integration/flags_test.go
@@ -173,4 +173,26 @@ var _ = Describe("Flags Specs", func() {
 		Ω(output).Should(ContainSubstring("0 Passed"))
 		Ω(output).Should(ContainSubstring("0 Failed"))
 	})
+
+	regextest := func(regexOption string, skipOrFocus string) string {
+		pathToTest = tmpPath("passing")
+		copyIn("passing_ginkgo_tests", pathToTest)
+		session := startGinkgo(pathToTest, regexOption, "--dryRun", "-v", skipOrFocus)
+		Eventually(session).Should(gexec.Exit(0))
+		return string(session.Out.Contents())
+	}
+
+	It("regexScansFilePath (enabled) should skip and focus on file names", func() {
+		output := regextest("-regexScansFilePath=true", "-skip=/passing/") // everything gets skipped (nothing runs)
+		Ω(output).Should(ContainSubstring("0 of 4 Specs"))
+		output = regextest("-regexScansFilePath=true", "-focus=/passing/") // everything gets focused (everything runs)
+		Ω(output).Should(ContainSubstring("4 of 4 Specs"))
+	})
+
+	It("regexScansFilePath (disabled) should not effect normal filtering", func() {
+		output := regextest("-regexScansFilePath=false", "-skip=/passing/") // nothing gets skipped (everything runs)
+		Ω(output).Should(ContainSubstring("4 of 4 Specs"))
+		output = regextest("-regexScansFilePath=false", "-focus=/passing/") // nothing gets focused (nothing runs)
+		Ω(output).Should(ContainSubstring("0 of 4 Specs"))
+	})
 })
--- a/internal/spec/specs.go
+++ b/internal/spec/specs.go
@@ -10,6 +10,7 @@ type Specs struct {
 	specs                 []*Spec
 	numberOfOriginalSpecs int
 	hasProgrammaticFocus  bool
+	RegexScansFilePath    bool
 }
 
 func NewSpecs(specs []*Spec) *Specs {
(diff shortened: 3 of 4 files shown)


jayunit100:
cc @robdimsdale yes, this is the PR we need.  thanks for the feedback, I did my best to provide context for why this patch is so critical for us in that issue.

jayunit100:
bump :)

james-lawrence:
what are you trying to solve that just using FIt functions wouldn't give you?

jayunit100:
@james-lawrence  we are filtering on source path, so that any test suite has a notion of namespacing that is built in when doing  extensions of large ginkgo suites.  `Fit()` Has the same semantics of `--focus`, meaning that it only is able to filter based on annotations which a user took the time to add to his spec.

james-lawrence:
I'm going to need a particular example, you're trying to skip specific tests inside of a single test suite outside of a development environment? sample code would be nice.

generally I just use standard linux CLI tools to filter packages I don't want to run (or want to run):
```
ginkgo $(go list github.com/onsi/ginkgo/... | grep -v package-to-remove-from-test)
```

and when I need a particular test I just use FIt...

I'm just not sure I see the value in executing just particular tests in a package?

jayunit100:
@james-lawrence https://github.com/openshift/origin/blob/master/test/extended/util/test.go#L119

james-lawrence:
yup clarifies it, have you looked into how the examples I provided worked? the examples I gave on how to accomplish this work and solve that particular example from the ginkgo CLI. you're not using the CLI but you should be able to simulate how that example works higher up in your process. i.e.) figure out the tests suites that need to be run earlier and pass only those packages to the test suite runner.

jayunit100:
If we filter before building the binary, then we have to build different packages...  We can't really afford to maintain separate packages....

*To help expose further why this is not only good for our case, but good for ginkgo as a whole, we might consider comparing this with other languages.*

- in java (maven/gradle) you would use something like to filter `**/${testGroup}/*`. 

- similarly, in pytest , you would do something like `pytest x ....` where x is the suite.  

- In apache bigtop, we built our own  integration test framework, and found that many vendors jumped on board and adopted this type of namespacing readily, using a groovy/gradle DSL to do something similar - and these are used by several major vendors to test their particular distributions: https://github.com/apache/bigtop/tree/master/bigtop-tests/smoke-tests/ .

james-lawrence:
> So, I hope I've made the case somewhat clear by now for adding this very simple and flexible mechanism for file path based namespacing.

and as I've tried to show you, ginkgo already supports it..... that is exactly what ginkgo $(go list ...) is doing its generating a list of packages and passing them to ginkgo. ginkgo already supports this... the problem is with how you're invoking it.

james-lawrence:
Let me see if I can find where ginkgo uses the package list provided at the cli. might help you out.

jayunit100:
Specifically we want the binary produced via `go test -c` to directly support file based namespacing.  Sorry if I didn't clarify that in my diatribe above :) :) . 

Do you know of a way to build a binary which supports these filters?  If so I agree, ginkgo fully supports this.

james-lawrence:
maybe lets talk outside of the PR so it doesn't get any messier =)

jayunit100:
Sure! my email is jay@apache.org.

james-lawrence:
so did a big dig into what jayunit is trying to accomplish basically boils down to not that ginkgo (as a tool) can't support what they are asking. but basically the library has limited abilities to be configured on how it skips specs.

the CLI basically avoids the issue by compiling multiple binaries 1 per package as far as I can tell. which allows regex filtering at the package level.

if ginkgo wants to support customizing the spec skipping logic per suite, then I'd probably change ApplyFocus method to take an interface that operates on an []Spec. then implement defaults skipping logic for the CLI and allow jayunit to specify their own implementation prior to RunSpecs being invoked.

jayunit100:
so, are you saying that adding the filepath to focus is too coupled?  Theoretically I guess that's true but ... It certainly seems like the filepath is easy to filter on as is, and it won't effect anything since it's a specific flag?

james-lawrence:
I was saying if we are going to start treating the internal stuff as an api for a test suite runner core library which ginkgo, and your application are based off of then we should likely consider just breaking out the filtering to allow it to actual be flexible, not against this current PR in anyway given your constraints.

onsi has mentioned not wanting to treat the filepath as special in another issue. my suggestion avoids that issue and allows you guys to be unblocked.

jayunit100:
Gotcha and agree.  Thanks for taking the time to see why this is so critical to us.

@onsi , can we get some feedback on how to proceed, it would really be a valuable feature for us to get this in, and I'd be happy to help with refactoring broader interface based filtering for the future if help is needed.

robdimsdale:
@jayunit100 @james-lawrence apologies for the delayed response. We're pretty busy at the moment. Thanks for taking a look into the problem @james-lawrence - that's really appreciated.

This PR looks ok in the sense that it does what it set out to achieve, but I'm still unconvinced we're solving the right problem. However, I'm willing to accept that I don't see the problem space from the same perspective as you, and that skipping on files really is the best way for you to proceed.

The other question of whether the ginkgo internal API should be exposed is an interesting one. As a developer I generally like to see APIs supported in addition to CLIs and while I haven't yet found a requirement for this split in ginkgo, I've certainly seen it in other projects. It feels like the better engineering solution - providing extension points for external integrations.

However, as a maintainer of the project, turning the internal API into an external one to bypass the CLI increases the surface area considerably, and I'm not sure we have the bandwidth to support that decision.

I essentially see four paths forward:

1. Merge this PR (or something very similar)
1. Investigate an alternative, proposed by @james-lawrence, which involves opening up the surface area of the project
1. Find an alternative to the above proposals
1. Reject any solutions

I really want to find a way to enable ginkgo to meet the needs of the community, and it certainly sounds like this is causing pain for the OpenShift OSS developers. Therefore I really don't want to close this PR and the associated issue with no resolution.

There's been some investigation into alternative solutions, and realistically I don't think we will see a better proposal than either of the two suggested so far.

Of the two remaining options, I think I'd probably prefer to see a PR skipping files by regex than commit the maintainers of this project to increasing the surface area by exposing the library.

@onsi do you have any thoughts here?

onsi:
Sorry to be so uncommunicative recently.

@jayunit100 I really do hear the pain y'all are in...

I'm really *not* a fan of this file-name based filtering as the package is the more fundamental unit in Go and coupling file names to tests in this way really feels like a smell.  I'm concerned that a consensus problem in one community is driving a solution into Ginkgo that, when looked at on its own merits doesn't make a lot of sense.

Let's step back.  Basically, there's a desire to:
- be able to group different tests by some sort of theme
- be able to focus/skip by thematic grouping

There are many ways to do this today:
- agree on a convention for tagging specs.  I've already proposed `Describe("{TAG1} {TAG2} Foo")` as a best practice that works well on the projects I've seen.  You can then use `--focus` and `--skip`.
- *add your own flags* to the test binary.  You can, in an `init` (I believe) add custom flags to your test.   You could then use these flags to control which tests to include or not include.  This would be an *explicit* thing that appears in your code and wouldn't be *implicitly* coupled to the name of the file the code happens to be in.
- use build tags and build different binaries with different build tags

It's really unclear to me why the `{TAG}` idea has been rejected by the k8s community.

There are a variety of changes and/or improvements we could make to Ginkgo:
- Add formal tag support to the DSL and CLI.  I'd be happy to help design this if folks think it will gain consensus and usage in the k8s community.
- Filter by filename - the proposed solution that I'm struggling with.

Sigh... I'm close to just letting this in.  I don't feel strongly enough that keeping this out of Ginkgo is worth y'all's suffering.

But I'd like to ask you to first please sincerely consider one the other alternatives I've listed above and explain to me why they won't work for you.  Is that fair?

jayunit100:
Thanks! Well, We want what's best for ginkgo too...We don't want to merge this if it makes ginkgo worse.  

1) So I will respond to your comments but lets NOT merge this if you think it's going to be a bad omen.  

-  golang packages don't compose easily as dependencies, it makes it impossible to ship a testing artifact.
-  tag conventions can't really be relied on at full scale.  And DSL wrappers are hard to enforce .
- run time skips aren't detected by dryRun, so adding just in time filters makes using a massive upstream test suite predictably imossible 

2) Final Thoughts

I very respectfully differ On the statement that "directory names a dangerous thing to filter on"... So if we still don't like this PR, then *Let's talk about a long term workaround that beat file filters* without bloating ginkgo.

- support a global wrapper interface that forces all ginkgo calls to go through a dryRun friendly Describe().
- support injection of "BeforeDryRun" so that tests can broadcast their status to the filters before runtime.  That allows one to add file name filtering or other complex logic as an app specifc feature.

jayunit100:
tl;dr lets do whats right for ginkgo - if this isn't the right thing, im happy to close, and we'll collaborate on a different feature which provides the same high level control namespacing.

smarterclayton:
I can speak to some of Jay's concerns (being from both Kube and OpenShift)

1. run time skips aren't detected by dryRun, so adding just in time filters makes using a massive upstream test suite predictably impossible

This is the ugliest part about filters.  We always have one or two tests that we need to exclude which aren't tagged correctly, or which fail for some reason, but a regex is hard to identify or target.  Being able to do programatic filters that could correctly remove things from skip and focus before the list is evaluated would be helpful.  Sometimes tests come in via code dependencies which we don't want to expose to the binary consumers (i.e. so a tester doesn't always have to add a skip that they don't care about).

2. tag conventions can't really be relied on at full scale

This is something that is hard enough to do in one project, let alone two.  Kube relies very heavily on a very large suite, which will double or triple in size over the next year.  OpenShift relies on that suite, plus our own growing suite, and we can fix / identify issues at different rates.  Regexes to filter tags isn't a great solution at very large number of tests if you're trying to subdivide based on criteria that are specific to your project, not necessarily the test author.

It's possible that inherited suites are a pattern that Ginkgo isn't designed or intended to solve - part of the reason we do that is sheer size of binary (Go 1.6 takes ~1-2 minutes to compile the test suite on 16 cores), but also so that we can offer a single unified tool for layered suites.  If we need to consider clearly delineating our suites by binary we can certainly do that.

jayunit100:
making binaries per suite untenable in the long run (1) complex godeps (2) complex packaging (3) complex use case for QE and end user.  

leading full circle to the motivation for this PR :)

jayunit100:
After taking a few days off @onsi , I think this is the right thing for ginkgo.
Now that the use case has been verified... Could I make a final appeal to merge it ? 
Or would you want me to close it ?
I'll leave the final call up to you :).

onsi:
I'll merge it in.  Please:

1. Add an integration test for Ginkgo.  There's a pile of tests around focussing/skipping.  Should be pretty self-explanatory but lemme know if you need help.
2. Update the documentation in the `gh-pages` branch and submit a PR for that too.

jayunit100:
Ok will do. I will have a PR for integration testing up by tomorrow

jayunit100:
Ok @onsi , 

- Added tests for the focus/skip combinations of regexScansFilePath `go test --ginkgo.focus="regexScansFilePath"`

- Updated the github.io docs in a separate PR. 

Should be all set.  Let me know if any other thoughts !

jayunit100:
ping @onsi  :)

jayunit100:
@onsi @robdimsdale all-systems-go to merge?

onsi:
so sorry, jayunit100 -- looking at this now..

PR #241, opened 2016-03-26
timeout update

jayunit100:
Makes the varargs functionality explicit @robdimsdale .
Where should I add a unit test for this?


jayunit100:
It doesn't seem like there are 
tests for the DSL, maybe i could create a `dsl_test.go` file?

onsi:
Hey @jayunit100 I appreciate you are trying to make the DSL a bit more explicit.  Truth be told, Ginkgo and Gomega play the varargs trick all over the place and I haven't seen anyone get tripped up by this so I'm inclined to leave it be.

Hope that's OK.

jayunit100:
I have gotten tripped up by it, so now you have seen one person...  But fair enough.... It certainly won't trip me up again :):):)

Issue #249, opened 2016-04-05
Proposal: Skip the `vendor` directory by default when using go 1.6+


sykesm:
When dependencies are included, `ginkgo -r` tries to test them too.  We're using `ginkgo -r --skipPackage ./vendor` to avoid this behavior today.

Happy to create a PR if you're interested in this.

Thanks.
@sykesm & @rosenhouse

robdimsdale:
@sykesm @rosenhouse check your version of ginkgo executable is up to date - this issue was raised and fixed in https://github.com/onsi/ginkgo/issues/187 (and again in https://github.com/onsi/ginkgo/issues/211) and should be fixed in newer versions of ginkgo.

Thanks,
Rob

sykesm:
The [code] is only looking at the `GO15VENDOREXPERIMENT` environment variable and isn't taking the go version into account. With go 1.6, the vendor support is enabled by default.

Thanks.

[code]: https://github.com/onsi/ginkgo/blob/master/ginkgo/testsuite/test_suite.go#L52-L58

robdimsdale:
Are you saying that you are using golang 1.6 and not observing the desired behavior by default (i.e. ginkgo is not skipping `./vendor`). If so, a PR would be welcome.

onsi:
Yep, would love to have a PR for this.  I believe the original PR left the
1.6 case dangling.  We should circle back on that.  Thanks @sykesm!

james-lawrence:
this is resolved with #253 being merged.

robdimsdale:
@james-lawrence thanks for the update - closing this issue now. @sykesm feel free to re-open if you see anything untoward.

PR #252, opened 2016-04-08
fix vendor for 1.6.

james-lawrence:
possible fix for #249.

this should still work for <1.5 when it set to 0 or 1, but when it is unset it would skip the vendor directory.

we could also check the go version command output but I'm not sure of every previous versions output I think this is easier.


onsi:
looks like you meant to kill this in favor of #253 ?

james-lawrence:
yes left it open if you had a preference on which route to go =)

PR #253, opened 2016-04-08
Fix go16 vendor v2

james-lawrence:
alternative fix for #249.

this should work across the board. updated primary tests.

I was a little heavy handed on the tests, read copy and paste, if someone wants to clean up that portion and scope them down to just the functions that'd be awesome.

### DIFF
--- a/ginkgo/testsuite/test_suite.go
+++ b/ginkgo/testsuite/test_suite.go
@@ -52,8 +52,10 @@ func SuitesInDir(dir string, recurse bool) []TestSuite {
 	// "This change will only be enabled if the go command is run with
 	// GO15VENDOREXPERIMENT=1 in its environment."
 	// c.f. the vendor-experiment proposal https://goo.gl/2ucMeC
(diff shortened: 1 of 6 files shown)


onsi:
What happens when go1.7 comes out?  Is there a way to make the build flags accept a minimum-version?

james-lawrence:
from go build documentation:
```
During a particular build, the following words are satisfied:

- the target operating system, as spelled by runtime.GOOS
- the target architecture, as spelled by runtime.GOARCH
- the compiler being used, either "gc" or "gccgo"
- "cgo", if ctxt.CgoEnabled is true
- "go1.1", from Go version 1.1 onward
- "go1.2", from Go version 1.2 onward
- "go1.3", from Go version 1.3 onward
- "go1.4", from Go version 1.4 onward
- "go1.5", from Go version 1.5 onward
- "go1.6", from Go version 1.6 onward
- any additional words listed in ctxt.BuildTags
```

onsi:
Gotcha - thanks!

Issue #260, opened 2016-06-20
Flake mitigation


lavalamp:
My goal here is to allow some grace around flaky tests. Unfortunately, they are a fact of life for the Kubernetes project. Sometimes that's due to programmer error and sometimes due to downstream dependencies, etc--I'm not arguing that I want flaky tests, just that I need to my CI system to understand the difference between "fails occasionally" and "totally broken". So I want to implement a mode that does something like this:

You'd call something like `ginkgo ... --test-tries=2 --test-required-passes=1 ...`. This tells ginkgo that each test is allowed two attempts and must pass at least once. Then ginkgo, after it runs a spec that fails, would run it again. If it passes the second time, then the entire suite is considered to have passed, but the individual failure is still reported in the JUnit output.

Does this seem like a patch that would be accepted?

lavalamp:
@onsi I'm looking at implementing this. Do you have suggestions on how to present this to the reporter.Reporter interface?

tedsuo:
As an alternative, what do you think about reporting that allows you to consume the failed tests, and re-run them yourself? Would that be too onerous?

The approach you're proposing is not without merit. If we did it that way, i think I would also want a different signal on exit, so CI systems can indicate that the tests only "passed".

lavalamp:
I want to rerun without having to go through BeforeSuite/AfterSuite. I
think what I'm proposing will still report the failures in e.g. JUnit while
claiming that the suite passed. I need that, anyway, to file issues about
the failures.

Amit-PivotalLabs:
@lavalamp:

> I want to rerun without having to go through BeforeSuite/AfterSuite.

Fair point.

@tedsuo:

> If we did it that way, i think I would also want a different signal on exit, so CI systems can indicate that the tests only "passed".

I believe most CI systems only know passed and failed, and won't know a different exit code means scare-quotes-"passed".  I imagine if don't want test runs with retried flakes in them to count as passing builds, you simply wouldn't use the flag in your build.  I'd be curious to hear @ryanmoran's thoughts on this proposal.

glyn:
Stipulating retries at the level of the whole build is too risky from my perspective. What if true non-deterministic bugs creep in? Then the build would likely still pass. Wouldn't it be better to identify specific flaky tests as such and supply a per-test retry count (as well as a string which could point to an issue tracker bug, for instance)?

(I wonder if @Amit-PivotalLabs wanted to hear @rmorgan's thoughts?)

[...]

PR #261, opened 2016-06-22
Add flaky test mitigation

lavalamp:
Implements #260

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -23,18 +23,17 @@ import (
 const VERSION = "1.2.0"
 
 type GinkgoConfigType struct {
-	RandomSeed          int64
-	RandomizeAllSpecs   bool
-	RegexScansFilePath  bool
-	FocusString         string
-	SkipString          string
-	SkipMeasurements    bool
-	FailOnPending       bool
-	FailFast            bool
-	FlakeAttempts       int
-	FlakePassesRequired int
-	EmitSpecProgress    bool
-	DryRun              bool
+	RandomSeed         int64
+	RandomizeAllSpecs  bool
+	RegexScansFilePath bool
+	FocusString        string
+	SkipString         string
+	SkipMeasurements   bool
+	FailOnPending      bool
+	FailFast           bool
+	FlakeAttempts      int
+	EmitSpecProgress   bool
+	DryRun             bool
 
 	ParallelNode  int
 	ParallelTotal int
@@ -78,7 +77,6 @@ func Flags(flagSet *flag.FlagSet, prefix string, includeParallelFlags bool) {
 	flagSet.BoolVar(&(GinkgoConfig.RegexScansFilePath), prefix+"regexScansFilePath", false, "If set, ginkgo regex matching also will look at the file path (code location).")
 
 	flagSet.IntVar(&(GinkgoConfig.FlakeAttempts), prefix+"flakeAttempts", 1, "Make up to this many attempts to run each spec.")
-	flagSet.IntVar(&(GinkgoConfig.FlakePassesRequired), prefix+"flakePassesRequired", 1, "Each test must pass this many times to be considered non-flaky.")
 
 	flagSet.BoolVar(&(GinkgoConfig.EmitSpecProgress), prefix+"progress", false, "If set, ginkgo will emit progress information as each spec runs to the GinkgoWriter.")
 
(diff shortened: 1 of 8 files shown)


tedsuo:
I see no integration tests for this feature, and I suspect that these new flags are not actually propagated. Can you add a test to `integration/flags_test` to prove that this works?

lavalamp:
OK, I'll add a test asap.

lavalamp:
Integration tests added. @tedsuo was there some particular thing you were concerned about? It passed as-is.

tedsuo:
At a glance, it looked like these flags would have to propagate through `BuildFlagArgs` in order to work, at least in some cases: https://github.com/lavalamp/ginkgo/blob/cd831bb9a6e6a46115ca953b13b4a5b61d2c2ca1/config/config.go#L100
https://github.com/onsi/ginkgo/blob/master/ginkgo/testrunner/test_runner.go#L226

Thanks for the integration tests. I'll manually review tomorrow.

lavalamp:
Ah, I didn't see that section. I will add them there tomorrow, that looks
important.

lavalamp:
@tedsuo I've added to that function and filed #263.

lavalamp:
@tedsuo sorry to be pushy-- any more comments?

onsi:
hi @lavalamp - sorry for the delay.

I'll commit to merging this in, but some comments:

- let's remove `FlakePassesRequired` - I'm not completely convinced of the usecase and would prefer to see it emerge from real-world usage

- I think we're missing some messaging to convey "this test has failed and is being retried".  Doing this well may entail changing the spec summary object or possibly the reporter interface itself.  As it stands I'm concerned folks will simply see the test suite *pass* but some tests look like they've failed.  I won't gate the PR on this but if you could spend a few minutes taking a look at the reporter/stenographer and consider whether there's a simple way to convey this context I'd appreciate it.

lavalamp:
> remove FlakePassesRequired

Done-- I was thinking I might want to require more than a 50% pass rate but we can cross that bridge when it becomes an actual problem.

> changing the spec summary object or possibly the reporter interface

I've run out of steam tonight but I'll try to take a look at this soon.

lavalamp:
I guess the obvious thing to do is add a "NumberOfFlakySpecs" to the suite summary and let the reporters handle that as appropriate.

The stenographer interface is harder, since a "flake" is really N failures followed by a pass for the same spec. I'll think about this more tomorrow.

At the very least I can add some documentation warning users about the behavior of this flag (pushed just now).

lavalamp:
I added the flake concept to the suite summary and to the spec object.

Please let me know if you think this is good, or if I should do something like run the test N times and then report the result a single time.

lavalamp:
@onsi anything I can do to move this along?

onsi:
Hey @lavalamp bear with me.  We're exploring a couple of different patterns for doing this.  I have a commit on a branch:

https://github.com/onsi/ginkgo/tree/runlists

That adds the ability to emit failed tests to a file then run only tests in said file.  Details:

ginkgo --writeFailuresFile=path/to/file.json

will write to file.json if and only if there are test failures.  it will export a list of failed tests detailing the name of the test, the coordinates of the test, and the failure reason.  if nothing fails no file is written however if a file with that name already exists it will not be removed (should it be?)

ginkgo --runFailuresFile=path/to/file.json

will run only the tests listed in file.json.  if file.json doesn't exist or is malformed ginkgo panics (so its on you to write the code that says "is this file here?  should i rerun the tests?). i think passing both flags in will work so you can rerun until there are no failures.

We're going to try this interface IRL and weigh it against your approach.  This requires a bit more work on the test runners side (e.g. a bash script that reruns ginkgo until it exits 0) but it does leave a better trail of breadcrumbs and will be harder for folks to abuse.

lavalamp:
That approach will call before/afterSuite multiple times. Also, putting
logic in bash is unfortunate. Maybe I can look into using the logic in this
change but emitting a second JUnit file for reruns?

tedsuo:
FWIW, there are ways to wed the two approaches without any extra mechanisms:

```bash
ginkgo --writeReport=path/to/file.json
ginkgo --writeReport=path/to/file.json --retryFailures=2
ginkgo --retryReport=path/to/file.json --retryFailures=1
```

#### ginkgo --writeReport=path/to/file.json [[--retryFailures=0]]
`writeReport` writes a Ginkgo Report File just before calling `AfterSuite`. It contains test telemetry. If ginkgo exits with success, the report contains details of a successful pass. If any tests fail, the report records the reason and trace for each failure. 

If `retryFailures` is set to a positive integer, failures in the report are then consumed.

#### ginkgo --retryReport=path/to/file.json [[--retryFailures=0]]
`retryReport` loads a Ginkgo Report File. If any failures were recorded, `BeforeSuite` is called,  and the recorded failures are rerun according to the behavior of `retryFailures`.

lavalamp:
I guess I don't understand what that gains us-- can't I just do that now by parsing the junit file and rerunning the failures?

onsi:
hey @lavalamp am going to pull this in.  so you mind writing up some docs for the gh-pages branch when you have a moment?

lavalamp:
Thanks! I will send you some docs.

Issue #176, opened 2015-08-10
Question: Modifying benchmark output to show milli/micro/nano seconds


duskhacker:
I'm interested in seeing my benchmark output reported in milliseconds instead of seconds. I've looked somewhat at the code, but there doesn't seem to be an easy way to do this without modifying the internal source. I've looked at writing a reporter, but that doesn't seem to do what I want, I seem to be able to write a reporter that outputs the actual measurement in the units I want, but I don't see a way of modifying the label 's' -> 'ms'. 

Is this possible, and if yes, could I get some pointers on how to do this without much fuss?

onsi:
hey @duskhacker sorry for the delay.

Unfortunately the `Measure` code is fairly opinionated (too opinionated) around how it records and represents time.

One approach could be to make it possible to specify the units and a precision for a benchmark:

1. Add a new method to the [`Benchmarker` interface](https://github.com/onsi/ginkgo/blob/master/ginkgo_dsl.go#L175-L178) called

```
RecordValueWithPrecision(name string, value float64, units string, precision int, info ...interface{})
```

2. Add `Precision int` to [`types.SpecMeasurement`](https://github.com/onsi/ginkgo/blob/master/types/types.go#L87-L103)

3. Update [`getMeasurement`](https://github.com/onsi/ginkgo/blob/master/internal/leafnodes/benchmarker.go#L44-L67) to accept and pass in the `FormatString`.  Add `RecordValueWithPrecision` and update `Time` and `RecordValue` to insert the existing precision (`3`).

4. Update [measurementReport](https://github.com/onsi/ginkgo/blob/master/reporters/stenographer/stenographer.go#L487-L542) to honor the measurement's `Precision` instead of the hardcoded `("%.3f")`.

---

This way you can `RecordValueWithPrecision("foo", duration.Milliseconds(), "ms", 2)  (note you'll have to do the timing yourself).

You probably don't need the `Precision` piece so you could just do `RecordValueWithUnits(name, value, units)` but I think `Precision` could be useful in general.

Final alternative: add a new `TimeWithPrecision()` method to `Benchmarker`.  Maybe this could take "s", "ms", "µs", etc.. maybe it just takes an integer precision.  Not sure but happy to think it through with you if you'd like.

onsi:
Closing this out for now.  @duskhacker lemme know if you'd like to work on this/reopen the issue.

PR #266, opened 2016-07-13
Allow units and precision in benchmark

jveski:
Addresses issue #176

This change adds support for benchmark values of varying units by adding a RecordValueWithPrecision method.


jveski:
@onsi you mentioned a TimeWithPrecision function in #176 - let me know if you’d like me to implement it here

onsi:
This is great!  thanks!

if you are up for the TimeWithPrecision work go for it - though I don't think it's strictly necessry

cobalt77:
@onsi Was just looking for a TimeWithPrecision function and found this thread, have you taken a look at #732 above? Would be excellent to have this available!

onsi:
i'm actually planning on deprecating the benchmark behavior as part of the [v2](#711) work currently in flight.  my sense is it's not widely used, adds complexity the codebase, and doesn't need to be part of core ginkgo (probably better off as a separate library that is focused on benchmarking and grows more featureful with time.)  With that said if you and/or others are using it extensively I'd love to learn more about the usecase/value it's bringing and how painful removing it would be.

cobalt77:
Since I am using ginkgo for all my tests the measure functionality integrates nicely where needed. it helps to be able to replace an It() with a Measure() here and there to be able to benchmark some sections without having to rewrite code or move it elsewhere. I do agree that turning it into a separate library would be an improvement though.

onsi:
any chance you could point me at a repo that has an example of your usage?
 just want to learn more from real-life examples.

onsi:
as i think about replacing the functionality - i'm leaning towards adding a new Gomega subpackage that specializes in measurements/benchmarking and is less tightly coupled to Ginkgo.  that would give folks a relatively straightforward migration path and resolve my main concerns with it being in Ginkgo core.  could you see that working for your usecase?

cobalt77:
The main project I am working on is private (unfinished) but I will get back to you this weekend with a gist link to some sample code. 
I think a gomega subpackage is a good idea. A drop-in replacement for my use case would be something like 
```go
It("Should do the thing", func() {
  Measure("the thing, func() {
    // some code
  }, iterations)
})
```
Whereas before my code would look like
```go
Measure("Doing the thing", func(b Benchmarker) {
  b.Time("the thing", func() {
    // some code
  })
}, iterations)
```
This could let you have matchers like `Expect(...).To(TakeOnAverage("<", 1*time.Second))` or something.

onsi:
Thanks @cobalt77 !  yes I'd imagine something along the lines of what you're describing along with a suite of associated matchers.

A couple more questions as I think about the user experience:
- do you find the benchmark report that Ginkgo emits to be useful?  do you actually look at it when tests run?
- or is it more valuable to analyze the results and make assertions like `TakeOnAverage("<"...)`? and only then see the results?
- have you generally used timing benchmarks?  Or do you also use [`RecordValue()`](https://github.com/onsi/ginkgo/blob/master/ginkgo_dsl.go#L193)?

cobalt77:
I've also used Time() to check how long Eventually() takes to complete. Here's an actual code sample:
```go
b.Time("Handling provider add callbacks", func() {
	testCh := make(chan struct{}, len(providers))
	// OnProviderAdded registers a callback (non blocking)
	l.OnProviderAdded(func(ctx context.Context, uuid string) {
		testCh <- struct{}{}
	})
	Eventually(func() int {
		return len(testCh)
	}, 1*time.Second, 1*time.Millisecond).Should(Equal(len(providers)))
})
```
This would bench how long it took for OnProviderAdded to be called len(providers) times.

onsi:
Thanks for all this detail!  You’ve given me some ideas for how to design a benchmarking module for gomega.  I’ll write those ideas down then circle back to get some input on them :)

onsi:
hey @kralicky if you're interested I've put together a beta version of a new benchmarking package in Gomega.  It's called [gmeasure](https://github.com/onsi/gomega/tree/master/gmeasure) and is in the latest Gomega release.  The package is documented but I'm working on additional documentation with richer examples for the onsi.github.io/gomega docsite.  Let me know if you have any thoughts/feedback.  It's a slightly different mental model but I think will provide a ton more flexibility going forward.

kralicky:
I will definitely check it out and let you know!

onsi:
thanks!  i'll let you know when the docs are finished too :)

onsi:
docs are up now at https://onsi.github.io/gomega/#gmeasure-benchmarking-code

kralicky:
@onsi Tried out gmeasure a bit [here](https://github.com/rancher/opni/blob/operator/test/e2e/demo_test.go#L221), looks pretty good so far! One thing I would recommend would be a way to control the sample rate in the SamplingConfig. Looks like it just samples as fast as possible right now. My intuition was that if you specified both a sample count and a duration, it would perform N samples in the given duration, spacing them out evenly.

onsi:
hey @kralicky - super cool to see how you're using this!  Looks great and please keep the feedback coming!

I'll add a `MinSamplingInterval` to the `SamplingConfig` - if a sample takes less time than `MinSamplingInterval` then `gmeasure` will wait until at least `MinSamplingInterval` has elapsed before collecting the next sample; otherwise the next sample is taken immediately.

The sample count and duration are intended to be maximums so you can say stuff like "do this 10 times or for up to 1 minute - whichever comes first"

PR #272, opened 2016-07-21
Use the go1.5 build tag to handle vendor exceptions

fsouza:
In this case Go 1.5 is the exception, and not the rule, so we should use
its build tag to determine the behavior. Otherwise, we will have to
extend the +build line to include all new Go releases in the future.

Related to #249 and #253.

### DIFF
--- a/ginkgo/testsuite/testsuite_test.go
+++ b/ginkgo/testsuite/testsuite_test.go
@@ -1,4 +1,4 @@
-// +build go1.6
+// +build !go1.5
 
 package testsuite_test
 
--- a/ginkgo/testsuite/vendor_check.go
+++ b/ginkgo/testsuite/vendor_check.go
@@ -1,4 +1,4 @@
-// +build go1.6
+// +build !go1.5
 
 package testsuite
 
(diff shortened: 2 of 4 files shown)


robdimsdale:
@fsouza thanks for the PR. What you say is true, and this PR looks fine, although I would point out that it's probably not actually an issue as when go 1.7 comes out ginkgo can drop support for golang 1.5, as there's little value in supporting versions that the golang group don't themselves support (i.e. they support the latest and the previous minor versions).

Still, I'd be happy to merge this PR. @onsi any objections?

onsi:
ship it!






On Wed, Jul 20, 2016 at 8:37 PM -0500, "Rob Dimsdale-Zucker" <notifications@github.com> wrote:












@fsouza thanks for the PR. What you say is true, and this PR looks fine, although I would point out that it's probably not actually an issue as when go 1.7 comes out ginkgo can drop support for golang 1.5, as there's little value in supporting versions that the golang group don't themselves support (i.e. they support the latest and the previous minor versions).



Still, I'd be happy to merge this PR. @onsi any objections?

robdimsdale:
Merged, thanks!

fsouza:
Thank you for merging this quickly!

onsi:
heyo looks like we're going to need to pull this.  @fsouza open to your thoughts but it sounds like this commit wasn't necessary after all?

PR #275, opened 2016-07-21
not running vendored tests in 1.6

tinygrasshopper:


tinygrasshopper:
Oh, looks like this has already been reverted.

PR #281, opened 2016-08-11
Covermode flag (and reworked pass-through flags passing)

allenluce:
This consolidates Go flags, limiting the number of things passed into the test runner. It also removes the intermediary definitions of the various flags from the `RunWatchAndBuildCommandFlags` and `TestRunner` structs leaving them only in the `flags()` and CompileTo()` functions.

The `covermode` flag is added as a pass-through.


onsi:
THANK YOU.  Much better :)

allenluce:
No problem, and thanks for the quick merge!

Issue #277, opened 2016-07-29
Pass through -work to go build


squeedee:
We had a test polution that i suspected was removing the intermediary `go build` files. I wanted to run `ginkgo -r -work` to prove it, but that's not an option.

I admit that `go test -work ./...` get's me there, but theres a small but time-consuming cognitive dissonance switching from ginkgo's syntax to go test's packages spec.

onsi:
I think the flag gets passed in now but Ginkgo doesn't print out the compilation output.  Will fix.

onsi:
this is now fixed.

PR #282, opened 2016-08-12
Redo flags again, add a bunch of pass-throughs.

allenluce:
This is a proposal to support a bunch of new flags. It should take care of my issue (I need `coverprofile`) as well as #277. It simplifies things in that it puts all the pass-through options into a single map but makes other things more complex as it's a map of interface values. 

This adds a degenerate test for the new flags but keeps the old flag tests intact.

### DIFF
--- a/ginkgo/build_command.go
+++ b/ginkgo/build_command.go
@@ -46,7 +46,7 @@ func (r *SpecBuilder) BuildSpecs(args []string, additionalArgs []string) {
 
 	passed := true
 	for _, suite := range suites {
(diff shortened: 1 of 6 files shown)


onsi:
hmm.. looks like the coverage integration test is consistently broken by this commit.  could you take a look @allenluce ?

And thanks for solving the piecemeal go flags problem once and for all!

allenluce:
No worries! Although I wouldn't go as far as "once and for all." For a little while, at least. Looks like everything is working now, let me know how that looks.

onsi:
:) for a little while, indeed.  But definitely an improvement.  Pulling this in now, thanks.

allenluce:
Cool, thanks for the quick merge!

PR #284, opened 2016-08-17
Add package path prefix to compilation output path only if missing

darinkrauss:
With the latest version of Ginkgo and a recent version of Go (1.6.3), if tests fail to compile then the error message file paths include the directory path twice. This patch checks to see if the path already includes the directory path and, if so, does not add it a second time. 

Note: I'm not sure what "older versions of Go" do not support the "-o" option and require the function `fixCompilationOutput`, but if those older versions are no longer supported by Ginkgo, then this entire function and its use can probably go away. I'll let you make that call.


onsi:
thanks @darinkrauss

Issue #287, opened 2016-08-31
GinkgoT().Logf() doesn't mimic newline behavior of testing.T


lcoote:
Using the standard golang testing.T object, if I call 

```
        t.Logf("yarp %d", 1)
        t.Logf("yarp %d", 2)
        t.Logf("yarp %d", 3)
```

then I get the following output:

```
    testy_test.go:9: yarp 1
    testy_test.go:9: yarp 2
    testy_test.go:9: yarp 3
```

however, if I call Logf with the object returned by GinkgoT(), my loglines are not separated newlines.

```
    yarp 1yarp 2yap 3
```

PR #288, opened 2016-09-01
#287 Ensure Logf/Skipf insert newline characters

lcoote:
The standard golang testing.T object methods Logf/Skipf log output with newlines. This change makes the GinkgoT() proxy consistent with this behavior.


onsi:
Thanks!

PR #291, opened 2016-09-14
Include flake count in test summary

timstclair:
Previously, flakes could result in confusing messages like:

```
Summarizing 1 Failure:

[Fail] [k8s.io] Pods [It] should contain environment variables for services [Conformance] 
/var/lib/jenkins/workspace/node-pull-build-e2e-test/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/test/e2e/framework/util.go:2125

Ran 97 of 129 Specs in 562.705 seconds
SUCCESS! -- 97 Passed | 0 Failed | 0 Pending | 32 Skipped 

Ginkgo ran 1 suite in 9m25.151025459s
Test Suite Passed

Success Finished Test Suite on Host tmp-node-e2e-4e4fc36a-e2e-node-containervm-v20160321-image
```

By including the `| 1 Flaked |` message, it should be slightly less confusing why thatt failure wasn't counted.

Fixes https://github.com/kubernetes/kubernetes/issues/31802

/cc @lavalamp

### DIFF
--- a/ginkgo/testrunner/test_runner.go
+++ b/ginkgo/testrunner/test_runner.go
@@ -337,7 +337,7 @@ func (t *TestRunner) runParallelGinkgoSuite() RunResult {
 	writers := make([]*logWriter, t.numCPU)
 	reports := make([]*bytes.Buffer, t.numCPU)
 
-	stenographer := stenographer.New(!config.DefaultReporterConfig.NoColor)
+	stenographer := stenographer.New(!config.DefaultReporterConfig.NoColor, config.GinkgoConfig.FlakeAttempts > 1)
 	aggregator := remote.NewAggregator(t.numCPU, result, config.DefaultReporterConfig, stenographer)
 
 	server, err := remote.NewServer(t.numCPU)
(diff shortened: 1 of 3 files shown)


lavalamp:
LGTM, didn't realize this was so easy.

lavalamp:
However you may want to consider only printing this when the flag is set, or maybe only when there actually are flakes.

onsi:
Could we make emitting the `| X flakes` segment conditional on `X > 0`?  I'd rather not have `0 flakes` be part of the standard output since this is a fairly nno-standard feature that we shouldn't really be encouraging folks to use outside of very specific and thought-out contexts.

onsi:
Lol race condition.  @lavalamp got there first.  wdyt @timstclair?

timstclair:
Makes sense. That's actually how I implemented it originally, but then thought it my be confusing to omit it sometimes. I think controlling it based off the flag setting makes the most sense, I was just being lazy about piping the flag through :)

timstclair:
Ok, piped the flag through. PTAL.

onsi:
LGTM

Issue #286, opened 2016-08-20
Expose spec suite randomization seed to test code


jawnsy:
Hello maintainer,

Thanks for your work and for sharing this great library. We're just starting to play with ginkgo (with gomega and agouti) and things are great so far. I particularly like that specs are executed in randomized order by default, with a seed provided to enable repeatable test runs, as this was a similar technique we were exploring for randomization in our own codebase.

Please consider exposing this seed to test code (for example, the code that invokes `gk.RunSpecs`), as it would be convenient to have a single seed that controls all nondeterministic aspects of the test.

Jonathan

PS: I also noticed that the default behaviour is to use the current system time for the seed, and I wonder if it would be useful to combine this with other data, like the current process ID? I can't really think of any reason why the current scheme might be problematic, though.

onsi:
Hey @jawnsy - glad to hear you are enjoying Ginkgo, Gomega and Agouti!

You can actually access the seed from test code.  It's a bit ugly (globals... blech) but if you import the `github.com/onsi/ginkgo/config` package you can get to it via `config.GinkgoConfig.RandomSeed`

onsi:
and, yeah, I don't think there's any harm in using the current time.  Seemed like the simplest thing to do at the time.

jawnsy:
Hey @onsi, that's awesome, thanks for that info & your very quick reply, I'll give it a shot!

If you can share your thoughts on how to best modify the code to expose this, I'd be happy to work on a PR - it's the least we can do, as beneficiaries of your work :smile: Fair warning, I'm a relative newbie to Go, which is probably why I wasn't able to find that global myself :laughing:

I switched over some of our code to use `time.Now().UnixNano()` for nanosecond-resolution timestamps, which seems pretty low-risk, and would make seed collisions effectively impossible. I'd be happy to submit a PR if you'd like.

cc my much smarter colleague, @ironcladlou

onsi:
:) if you'd like you can add a `GinkgoRandomSeed` identical to `GinkgoParallelNode` here: 
https://github.com/onsi/ginkgo/blob/master/ginkgo_dsl.go#L71-L75

I'm not 100% convinced we need to worry too much about seed collision (what's the harm?) and the UnixNano numbers are pretty long.  Any strong reasons driving you towards it?

jawnsy:
[...]

PR #293, opened 2016-09-20
Expose the random seed via GinkgoRandomSeed()

jawnsy:
This change exposes the random seed used internally by Ginkgo to randomize
spec execution order.  If your own tests make use of randomness, using the
same seed ensures that your test executions are repeatable.  This can be
helpful for tests that randomly generate test data.

Closes: #286


jawnsy:
@onsi Hey, please take a look.  This was a pretty simple change (following your instructions on #286).  I wasn't quite sure how to add automated tests, though - please let me know how one would do that :)

onsi:
While we could add an integration test for this such a test would be pretty low value, IMO.  I'll add a unit test after I merge this in.

PR #298, opened 2016-09-28
Add syntax highlighting to docs

### LINKED ISSUE #220: docs pages should have syntax highlighting
Apparently github pages now support jekyll 3, and with it the `kramdown` syntax, coupled with `rouge` syntax highlighting. I'm curious as to whether this might enable the docs to have syntax highlighting - if so this could be helpful.

see: https://github.com/blog/2100-github-pages-now-faster-and-simpler-with-jekyll-3-0

Alexendoo:
Looks like: https://alexendoo.github.io/ginkgo/

Colours can be changed easily enough from now on, I went with just strings/comments for now since it seemed to fit well with the DSL

closes #220

### DIFF
--- a/css/syntax.css
+++ b/css/syntax.css
@@ -1,0 +1,8 @@
+.language-go .s {
+  color: #2dad6c;
+}
+
+.language-go .c {
+  color: #2dad6c;
+  font-style: italic;
+}
--- a/index.md
+++ b/index.md
@@ -2,7 +2,7 @@
 layout: default
 title: Ginkgo
 ---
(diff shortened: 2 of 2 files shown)


robdimsdale:
Looks great thanks!

robdimsdale:
Interestingly, we just received the following message from github:

```
The page build completed successfully, but returned the following warning:

You are attempting to use the 'pygments' highlighter, which is currently unsupported on GitHub Pages. Your site will use 'rouge' for highlighting instead. To suppress this warning, change the 'highlighter' value to 'rouge' in your '_config.yml' and ensure the 'pygments' key is unset. For more information, see https://help.github.com/articles/page-build-failed-config-file-error/#fixing-highlighting-errors.

For information on troubleshooting Jekyll see:

https://help.github.com/articles/troubleshooting-jekyll-builds
```

I don't think that's a consequence of this PR, just surfacing it for others to see.

Alexendoo:
Ah just checked my emails to find a couple of those as well, didn't think to check the config but yeah it's using rouge rather than pygments so it can just be renamed

PR #311, opened 2016-11-04
Fix tests for single node machine

alex-slynko:
When there is only one node, number of nodes is not shown and test fails.


onsi:
huh.  good catch.  thanks.

PR #318, opened 2016-12-24
Include captured output from failed tests into JUnit

smarterclayton:
JUnit has the ability to report the full output of a test run via the
system-out field which is not currently exposed. This commit turns the
test Writer in a capturing writer even when streaming so that the JUnit
reporter will have access to CapturedOutput from the Spec. Failed tests
include that output.

For suites that generate a lot of log output, being able to zoom in to a
particular failed test and see the output all at once makes various
JUnit visualization outputs more effective (like in Jenkins).


onsi:
looks like this broke one of the tests.  can you take a look @smarterclayton ?

smarterclayton:
Sorry about that, didn't adjust the test for the issue.  Fixed now.

onsi:
great thanks!

PR #319, opened 2017-01-20
Add the ability to use ./... to recursively test directories

tedkornish:
Existing `go` commands (`go test`, `go install`, `go build`, etc) let the user recurse into subpackages by ending a path, relative or absolute, with `/...`. Thought it would a nice usability enhancement. This also provides the ability to recurse into some directories and not others, which can be useful for finer-grain control when running expensive `ginkgo watch` processes.

e.g.

```
ginkgo watch ./only/this/package ./and/all/...
```

### DIFF
--- a/ginkgo/main.go
+++ b/ginkgo/main.go
@@ -234,7 +234,7 @@ func complainAndQuit(complaint string) {
 	os.Exit(1)
 }
 
-func findSuites(args []string, recurse bool, skipPackage string, allowPrecompiled bool) ([]testsuite.TestSuite, []string) {
+func findSuites(args []string, recurseForAll bool, skipPackage string, allowPrecompiled bool) ([]testsuite.TestSuite, []string) {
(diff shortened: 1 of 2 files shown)


ginkgo/main.go
```diff
@@ -246,10 +246,15 @@ func findSuites(args []string, recurse bool, skipPackage string, allowPrecompile
 					continue
 				}
 			}
-			suites = append(suites, testsuite.SuitesInDir(arg, recurse)...)
+			recurseForSuite := recurseForAll
+			if strings.HasSuffix(arg, "/...") && arg != "/..." {
```
tedkornish: Let me know how you feel about excluding `/...`. It seemed reasonable at the time but since Ginkgo does nothing to prevent running `ginkgo -r /` I suppose there's no reason to enforce that here either.

onsi:
This is great - thanks @tedkornish !

PR #322, opened 2017-02-16
Add warning about vendoring the table extension

georgeharley:


onsi:
thanks @georgeharley !

Issue #326, opened 2017-02-28
Don't print colorized output when not outputting to a TTY


stevekuznetsov:
With code like:

```go
import (
    "os"
    "golang.org/x/crypto/ssh/terminal"
)
 
func main() {
    return terminal.IsTerminal(int(os.Stdout.Fd()))
}
```

It should be possible to determine if the output from `ginkgo` is going to a TTY or not. It should be best practice not to spit out color escape codes if the output is _not_ a TTY.

robdimsdale:
I can see arguments for both cases. There are examples of where you want to output color to something that isn't a TTY - concourse.ci is one of these - so if we switched the default I would still expect to see a way to configure color.

Changing the default here could cause frustrations for these users as they would lose color support until they explicitly opted-in by re-enabling color.

stevekuznetsov:
I understand the backward compatibility argument, but given that it's a color and a flag `-color` could be used to force coloring, I think it's still true that well formed programs shouldn't output ANSI escape codes to file descriptors that aren't TTYs by default.

vito:
FWIW, Concourse runs its tasks with a TTY, so it would be unaffected by this. I agree that by default it should omit color. Curious if Travis CI and others would be affected though?

robdimsdale:
@vito I didn't realize that - was this always the case? I thought I've seen ANSI codes fail to be escaped in concourse, but I might be confusing it with something else.

A quick google search implies that Travis CI runs a TTY/PTY so that would likely continue to work. I vaguely remember GoCD not supporting ANSI color codes, but I'm not sure how many ginkgo users run on that system anymore.

@onsi thoughts on this potentially backwards-incompatible-but-probably-better feature request?

onsi:
Makes sense to me.  Steve are you up for submitting a PR?

stevekuznetsov:
Yes, I'll take a stab at it. Just to make sure we've got agreement on the change - if the output is not a TTY, act as if `-noColor` had been passed, _unless_ the user supplied `-color`.

vito:
yep. we initially did it for Bosh cli a long time ago.

manosnoam:
FYI,
When running in GoLand IDE, the -ginkgo.noColor removes colors from the terminal output as well:

![image](https://user-images.githubusercontent.com/9913945/63108640-0913d000-bf90-11e9-8050-91e7fbfb8f79.png)

PR #328, opened 2017-03-03
Don't colorize output by default if not writing to a TTY

stevekuznetsov:
The current behavior of `ginkgo` is to write ANSI escape codes to the
output stream to colorize messages by default. This leads to mangled and
unreadable logs unless the user knows to pass `--noColor`. Instead, the
default reporter should check to see if the output file descriptor is a
TTY and colorize output only if it is, or if the user has specifically
provides the `--color` flag.

This patch breaks backward compatibility for users that expected ANSI
escape codes printed to non-TTY file descriptors, but those users can
pass `--color` to return to the previous behavior. All other users
should not notice this change.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

Fixes https://github.com/onsi/ginkgo/issues/326#issuecomment-283958838
/cc @smarterclayton

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -15,9 +15,11 @@ package config
 
 import (
 	"flag"
-	"time"
-
 	"fmt"
+	"time"
+	"os"
+
+	"golang.org/x/crypto/ssh/terminal"
 )
 
 const VERSION = "1.2.0"
@@ -44,12 +46,37 @@ type GinkgoConfigType struct {
 var GinkgoConfig = GinkgoConfigType{}
 
 type DefaultReporterConfigType struct {
-	NoColor           bool
 	SlowSpecThreshold float64
 	NoisyPendings     bool
 	Succinct          bool
 	Verbose           bool
 	FullTrace         bool
+
+	color   bool
+	noColor bool
+
+	// Colorize caches whether or not we want to colorize output
+	Colorize *bool
+}
+
+// ColorizeOutput determines if output should be colorized. The user
+// can specify they want color (--color) or not (--noColor). Further-
+// more, we detect if the output to which we are printing is a TTY.
+// If the output is a TTY, the default is to use color but a user can
+// still request no color; if the output is not a TTY, the default is
+// to not use color but the user can still request it.
+func (c *DefaultReporterConfigType) ColorizeOutput() bool {
+	if c.Colorize == nil {
+		var colorize bool
+		if terminal.IsTerminal(int(os.Stdout.Fd())) {
+			colorize = !c.noColor
+		} else {
+			colorize = c.color
+		}
+		c.Colorize = &colorize
+	}
+
+	return *c.Colorize
 }
 
 var DefaultReporterConfig = DefaultReporterConfigType{}
@@ -87,7 +114,8 @@ func Flags(flagSet *flag.FlagSet, prefix string, includeParallelFlags bool) {
 		flagSet.StringVar(&(GinkgoConfig.StreamHost), prefix+"parallel.streamhost", "", "The address for the server that the running nodes should stream data to.")
 	}
 
-	flagSet.BoolVar(&(DefaultReporterConfig.NoColor), prefix+"noColor", false, "If set, suppress color output in default reporter.")
+	flagSet.BoolVar(&(DefaultReporterConfig.color), prefix+"color", false, "If set, enable color output in default reporter even when not printing to a TTY.")
+	flagSet.BoolVar(&(DefaultReporterConfig.noColor), prefix+"noColor", false, "If set, suppress color output in default reporter.")
 	flagSet.Float64Var(&(DefaultReporterConfig.SlowSpecThreshold), prefix+"slowSpecThreshold", 5.0, "(in seconds) Specs that take longer to run than this threshold are flagged as slow by the default reporter.")
 	flagSet.BoolVar(&(DefaultReporterConfig.NoisyPendings), prefix+"noisyPendings", true, "If set, default reporter will shout about pending tests.")
 	flagSet.BoolVar(&(DefaultReporterConfig.Verbose), prefix+"v", false, "If set, default reporter print out all specs as they begin.")
@@ -159,8 +187,8 @@ func BuildFlagArgs(prefix string, ginkgo GinkgoConfigType, reporter DefaultRepor
 		result = append(result, fmt.Sprintf("--%sregexScansFilePath", prefix))
 	}
 
-	if reporter.NoColor {
-		result = append(result, fmt.Sprintf("--%snoColor", prefix))
+	if reporter.ColorizeOutput() {
+		result = append(result, fmt.Sprintf("--%scolor", prefix))
 	}
 
 	if reporter.SlowSpecThreshold > 0 {
--- a/internal/remote/aggregator_test.go
+++ b/internal/remote/aggregator_test.go
@@ -32,8 +32,9 @@ var _ = Describe("Aggregator", func() {
 	)
 
 	BeforeEach(func() {
+		colorize := true
 		reporterConfig = config.DefaultReporterConfigType{
-			NoColor:           false,
+			Colorize:          &colorize,
 			SlowSpecThreshold: 0.1,
 			NoisyPendings:     true,
 			Succinct:          false,
--- a/ginkgo/notifications.go
+++ b/ginkgo/notifications.go
@@ -123,18 +123,18 @@ func (n *Notifier) RunCommand(suite testsuite.TestSuite, suitePassed bool) {
 		output, err := exec.Command(parts[0], parts[1:]...).CombinedOutput()
 		if err != nil {
 			fmt.Println("Post-suite command failed:")
-			if config.DefaultReporterConfig.NoColor {
-				fmt.Printf("\t%s\n", output)
-			} else {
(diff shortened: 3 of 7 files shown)


internal/remote/aggregator_test.go
```diff
@@ -32,8 +32,9 @@ var _ = Describe("Aggregator", func() {
 	)
 
 	BeforeEach(func() {
+		colorize := true
 		reporterConfig = config.DefaultReporterConfigType{
-			NoColor:           false,
+			Colorize:          &colorize,
```
stevekuznetsov: @onsi I made `DefaultReporterConfigType.Colorize` exported for these cases, but this leads to a confusing API for non-test users of the reporter. Please let me know if it is acceptable to either drop the explicit initialization in these tests (as it doesn't seem like the tests are testing whether the output is colorized or not) or if you want a different API here.

config/config.go
```diff
@@ -44,12 +46,37 @@ type GinkgoConfigType struct {
...
+// ColorizeOutput determines if output should be colorized. The user
+// can specify they want color (--color) or not (--noColor). Further-
+// more, we detect if the output to which we are printing is a TTY.
+// If the output is a TTY, the default is to use color but a user can
+// still request no color; if the output is not a TTY, the default is
+// to not use color but the user can still request it.
+func (c *DefaultReporterConfigType) ColorizeOutput() bool {
+	if c.Colorize == nil {
+		var colorize bool
+		if terminal.IsTerminal(int(os.Stdout.Fd())) {
```
stevekuznetsov: There's nothing that necessarily ties the default reporter to `stdout`, so technically this may be incorrect. I can refactor to accept to take the output stream as a parameter if that is more appropriate.

onsi:
thanks!

james-lawrence:
@onsi this is broken.

```4] flag provided but not defined: -ginkgo.color
[4] Usage of /tmp/ginkgo726661405/httputilx.test:
[4]   -ginkgo.dryRun
[4]     	If set, ginkgo will walk the test hierarchy without actually running anything.  Best paired with -v.
[4]   -ginkgo.failFast
[4]     	If set, ginkgo will stop running a test suite after a failure occurs.
[4]   -ginkgo.failOnPending
[4]     	If set, ginkgo will mark the test suite as failed if any specs are pending.
[4]   -ginkgo.focus string
[4]     	If set, ginkgo will only run specs that match this regular expression.
[4]   -ginkgo.noColor
[4]     	If set, suppress color output in default reporter.
[4]   -ginkgo.noisyPendings
[4]     	If set, default reporter will shout about pending tests. (default true)
[4]   -ginkgo.parallel.node int
[4]     	This worker node's (one-indexed) node number.  For running specs in parallel. (default 1)
[4]   -ginkgo.parallel.streamhost string
[4]     	The address for the server that the running nodes should stream data to.
[4]   -ginkgo.parallel.synchost string
[4]     	The address for the server that will synchronize the running nodes.
[4]   -ginkgo.parallel.total int
[4]     	The total number of worker nodes.  For running specs in parallel. (default 1)
[4]   -ginkgo.progress
[4]     	If set, ginkgo will emit progress information as each spec runs to the GinkgoWriter.
[4]   -ginkgo.randomizeAllSpecs
[4]     	If set, ginkgo will randomize all specs together.  By default, ginkgo only randomizes the top level Describe/Context groups.
[4]   -ginkgo.seed int
[4]     	The seed used to randomize the spec suite. (default 1488830923)
[4]   -ginkgo.skip string
[4]     	If set, ginkgo will only run specs that do not match this regular expression.
[4]   -ginkgo.skipMeasurements
[4]     	If set, ginkgo will skip any measurement specs.
[4]   -ginkgo.slowSpecThreshold float
[4]     	(in seconds) Specs that take longer to run than this threshold are flagged as slow by the default reporter. (default 5)
[4]   -ginkgo.succinct
[4]     	If set, default reporter prints out a very succinct report
[4]   -ginkgo.trace
[4]     	If set, default reporter prints out the full stack trace when a failure occurs
[4]   -ginkgo.v
[4]     	If set, default reporter print out all specs as they begin.
[4]   -test.bench regexp
[4]     	run only benchmarks matching regexp
[4]   -test.benchmem
[4]     	print memory allocations for benchmarks
[4]   -test.benchtime d
[4]     	run each benchmark for duration d (default 1s)
[4]   -test.blockprofile file
[4]     	write a goroutine blocking profile to file
[4]   -test.blockprofilerate rate
[4]     	set blocking profile rate (see runtime.SetBlockProfileRate) (default 1)
[4]   -test.count n
[4]     	run tests and benchmarks n times (default 1)
[4]   -test.coverprofile file
[4]     	write a coverage profile to file
[4]   -test.cpu list
[4]     	comma-separated list of cpu counts to run each test with
[4]   -test.cpuprofile file
[4]     	write a cpu profile to file
[4]   -test.memprofile file
[4]     	write a memory profile to file
[4]   -test.memprofilerate rate
[4]     	set memory profiling rate (see runtime.MemProfileRate)
[4]   -test.mutexprofile string
[4]     	write a mutex contention profile to the named file after execution
[4]   -test.mutexprofilefraction int
[4]     	if >= 0, calls runtime.SetMutexProfileFraction() (default 1)
[4]   -test.outputdir dir
[4]     	write profiles to dir
[4]   -test.parallel n
[4]     	run at most n tests in parallel (default 4)
[4]   -test.run regexp
[4]     	run only tests and examples matching regexp
[4]   -test.short
[4]     	run smaller test suite to save time
[4]   -test.timeout d
[4]     	fail test binary execution after duration d (0 means unlimited)
[4]   -test.trace file
[4]     	write an execution trace to file
[4]   -test.v
[4]     	verbose: print additional output````

**Comment by cscatolini** [comment]
@onsi the same problem as reported by @james-lawrence is breaking our CI.

**Comment by stevekuznetsov** [comment]
@james-lawrence @cscatolini what is the error in that output snippet and how would I address it? I also had some outstanding questions on the code here... wasn't quite expecting a merge this morning

**Comment by james-lawrence** [comment]
[4] flag provided but not defined: -ginkgo.color

**Comment by james-lawrence** [comment]
sorry didn't mean to strip the error out =)

File: config/config.go [hunk 3]
```diff
@@ -159,8 +187,8 @@ func BuildFlagArgs(prefix string, ginkgo GinkgoConfigType, reporter DefaultRepor
 		result = append(result, fmt.Sprintf("--%sregexScansFilePath", prefix))
 	}
 
-	if reporter.NoColor {
-		result = append(result, fmt.Sprintf("--%snoColor", prefix))
+	if reporter.ColorizeOutput() {
+		result = append(result, fmt.Sprintf("--%scolor", prefix))
```
stevekuznetsov: So do we want to change this to pass `--%snoColor` still? The migration strategy here would be that someone would need to compile their test with an updated `ginkgo` import to get the `--color` flag on their test binary, right?

james-lawrence:
@stevekuznetsov alright if I had to guess, we're missing setting up a flag somewhere. but not familiar with this section of the codebase.

onsi:
Sorry everyone. Reverted the commit.  Am pretty slammed today but will try to dig into this.  Apologies for the premature merge...

stevekuznetsov:
If you think the change in #332 is correct let me know and I'll collapse it all into a new PR.

onsi:
hey @stevekuznetsov - sorry finally finding bandwidth for Ginkgo.  If you're still up for collaborating on this we can try to figure this out together.

I don't think #332 will fix the issue.  The issue, as I understand it, is that PR #328 (i.e. this PR) deprecated `--noColor` in favor of `--color` which needs to be set explicitly to true when there is no TTY attached to the test process.

The issues is that `ginkgo build` makes it possible to build test binaries that can be run later by invoking `ginkgo` and pointing it at the test binary.  This means the version of the `ginkgo` CLI used to _run_ the tests need not be the same as the version o the Ginkgo library that was used to compile the tests.

#332 fixes the case where you are running a new Ginkgo CLI against a test compiled prior to introducing `--color` but I think it will actually break if you run the new Ginkgo CLI against a test compile _after_ introducing `--color` as it will pass `--noColor` to the test.

Our options are:

1.  Hold our noses and stick with `--noColor=false` to turn color back on for non-TTY terminals.
2. Introduce a `--color` as a new thing and print out a deprecation warning for `--noColor`, but continue to support it as well.
3. We can claim that the contract between the Ginkgo binary and compiled test suite version is not strong and that we only guarantee interoperability if both versions match and ask users to upgrade.

None of these options are great 👎 

Thoughts from anyone on this thread?

stevekuznetsov:
@onsi Hey -- thanks for having a moment to look at this.

> I think it will actually break if you run the new Ginkgo CLI against a test compile after introducing `--color` as it will pass `--noColor` to the test.

I am not sure that will be the case. Here's how I understand it -- today the Ginkgo runner passes `--noColor` to  a test binary to indicate what they want. #332  keeps that the same. If you use an old runner, you'll be passing `--noColor`. New test binaries still know what that flag means, so they will work as expected.

manosnoam:
Any update on this issue ?

Ginkgo v1.10.2 still prints ansi colors into junit reports (xml files), by default.
If specifying `--noColor` then both junit output and console output (for example on Travis-CI) does not include colors. 

Can we omit colors from reports, but keep colors in console (TTY/PTY) ?

PR #332, opened 2017-03-06
Pass `--noColor` to test binary for compatibility

stevekuznetsov:
If a user compiles their test binary with an older version of the Ginkgo
library, but uses a newer version of the Ginkgo test runner to run it,
they will not be able to pass the new flag `--color` to the test binary.
Instead, we should continue pasing the `--noColor` flag to preserve
compatibility.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

@onsi this may be the correct solution for the problems detected by @james-lawrence and  @cscatolini in https://github.com/onsi/ginkgo/pull/328

@james-lawrence,  @cscatolini could you try this patch set out to test?

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -187,8 +187,8 @@ func BuildFlagArgs(prefix string, ginkgo GinkgoConfigType, reporter DefaultRepor
 		result = append(result, fmt.Sprintf("--%sregexScansFilePath", prefix))
 	}
 
-	if reporter.ColorizeOutput() {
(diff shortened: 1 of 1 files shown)


james-lawrence:
this did resolve the issue for me. alternatively I think there we just need to ensure the compiled test program has the correct flag setup for the --color option. not sure how that is handled.

onsi:
Closing this one out in favor of the conversation on #328

Issue #305, opened 2016-10-12
Ginkgo "corrupts" some global state when testing "package main"


nikhilm:
I have a project that has certain functions in `package main` that I want to test. The tests are located in another directory. I am aware that this is bad style, and I should move the exported functions to another package and so on... but the fact is, it is something that works, until running ginkgo makes it not work.

I have a reduced test case at https://github.com/nikhilm/ginkgo-broken that intentionally _does not_ use any gomega specifics, to demonstrate the problem is in the `ginkgo` executable. Reproducing the README here.

Reproduced on 

    go version go1.7.1 darwin/amd64
    Darwin Nikhils-MBP 15.5.0 Darwin Kernel Version 15.5.0: Tue Apr 19 18:36:36 PDT 2016; root:xnu-3248.50.21~8/RELEASE_X86_64 x86_64
    Ginkgo Version 1.2.0

Environment:

    GOARCH="amd64"
    GOBIN=""
    GOEXE=""
    GOHOSTARCH="amd64"
    GOHOSTOS="darwin"
    GOOS="darwin"
    GOPATH="/Users/nikhil/iron/gocode"
    GORACE=""
    GOROOT="/usr/local/go"
    GOTOOLDIR="/usr/local/go/pkg/tool/darwin_amd64"
    CC="clang"
    GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/3t/tx0j5bws5kv7m6s2rj2dtyvm0000gn/T/go-build959139225=/tmp/go-build -gno-record-gcc-switches -fno-common"
    CXX="clang++"
    CGO_ENABLED="1"

## What is the problem?

When a `package main` exports some fields/functions that the test package
imports to test, ginkgo works fine on the first run, but fails on subsequent
runs. No global state seems to be modified, nor are new files created, so
I cannot figure out what is the cause. After the command is run, `go test`
starts failing too.

## Steps to Reproduce

    go get github.com/nikhilm/ginkgo-broken
    cd $GOPATH/src/github.com/nikhilm/ginkgo-broken
    cd tests
    go test -v . # everything works!
    go test -v . # everything works!

These commands output:

    === RUN   TestYo
    YOYOYO
    --- PASS: TestYo (0.00s)
    PASS
    ok      github.com/nikhilm/ginkgo-broken/tests  0.006s

Let's continue:

    ginkgo -v . # everything works!

[...]

PR #337, opened 2017-03-20
Fix testing suites outside of the current directory tree with Go 1.8 and test against Go 1.8 in CI

goonzoid:
This was a fun one to figure out! :clown_face:

Ginkgo's own integration suite was failing when using go 1.8. It turns out that a change in Go 1.8 means that relative import paths outside of the current directory tree can't begin with `./`.

Fix this by adding some logic to detect relative paths outside of the current directory tree, and then took the liberty of adding Go 1.8 to `.travis.yml`.


goonzoid:
And _of course_ it fails on CI 😢 

I'll take a closer look tomorrow or later this week and see if I can figure out what's different there compared to my local machine (may well be an OS X vs Linux thing I suppose).

goonzoid:
So, turns out the integration suite is significantly slower using Go 1.8 ☹️ 

I bumped the default `Eventually` timeout in the integration suite to 30s (from 15s) and it's green now. Not sure if that's the right thing, or if it would be better to dig in to the perf issues. If that's preferred, I can back out the timeout change and the commit that adds Go 1.8 testing to CI and that can be handled separately.

Let me know! 🍔

onsi:
Hmm.  That's concerning.  I think we should dig into the perf issues but
that's almost certainly a ton of scope creep for you Will.  I can try to
find some time to do this but if you're up for taking a first whack at it
I'd appreciate it.  TBH go compilations have seemed to get slower and
slower since 1.5... that may be more anecdotal than fact-based but i
suspect something like that is to blame here.

One quick thing to try: can you drop the `-i` arg on this line:

https://github.com/onsi/ginkgo/blob/master/ginkgo/testrunner/test_runner.go#L61

and try the tests again?  I've never fully understood how `-i` works and
whether it helps or hinders.  See this issue
https://github.com/onsi/ginkgo/issues/305 for what might be relevant
context is the `-i` thing helps/hinders.

goonzoid:
I've reverted the timeout increase and pushed a change without the `-i` flag just to see what happens on CI. I'm getting confusing and inconsistent results on my dev machine, but I'd like to dig in to it more in a day or two when I have some more time.

goonzoid:
@onsi looks like you've already figured this out on `master`, so I'll close this.

PR #340, opened 2017-04-03
Version was still at 1.2.0

pdelagrave:


onsi:
welp.  I think I need to do this as a 1.3.1 and cut a new release :/

onsi:
ok, fixed now.  cut a 1.3.1

PR #351, opened 2017-05-16
Fix wrong repo URL

baopham:


onsi:
durp - thank you!

esther-ng:
New URL doesn't seem to build binary.

When we `go get github.com/onsi/ginkgo`, we get the src for ginkgo, but it does not build the binary since the main does not exist in this directory.

We had to `go get github.com/onsi/ginkgo/ginkgo` to build the binary, which allow us to run the`ginkgo` command.

baopham:
Oh no, shoot. When I use the old URL, it complained so I thought it's wrong URL. We should revert this then.

onsi:
oh hey - i somehow got confused and misread the original PR.  yes you want to `go get github.com/onsi/ginkgo/ginkgo` as this will both fetch the source _and_ install the CLI.

PR #362, opened 2017-07-10
Add a flag to treat all test lambas as async, enabling global timeouts

lucas-clemente:


onsi:
Hi @lucas-clemente - sorry for the delay in getting back to you.

I have been long-meaning to deprecate the async functionality in Ginkgo in favor of the async support provided by Gomega's `Eventually` matcher.  Because it is impossible for one goroutine to terminate another Ginkgo cannot proactively kill asynchronous specs that hit a timeout.  This results in the specs running in the background and, potentially, interfering with subsequent tests in unexpected and confusing ways.  This doesn't happen often but when it _does_ happen it can be super flaky and super frustrating to debug.

You can learn more about Eventually [here](http://onsi.github.io/gomega/#making-asynchronous-assertions) - note that the argument to Eventually is expected to return immediately without blocking.

lucas-clemente:
Thanks for the reply. I agree that everything that can be done with async tests can equally be done with Eventually – and if it simplifies your code it might be worth deprecating duplicate functionality.

That being said, I don't fully understand your argument against async tests. If you have a blocking function (e.g. a Read()) and you want to test that it eventually returns, you can do that with async tests like this:

```go
It("works", func(done Done) {
  n, err := r.Read()
  // Assertions
  close(done)
}, 1)
```

With Eventually, you have to write it like this:

```go
It("works", func() {
  var done int32
  var n int
  var err error
  go func() {
    n, err = r.Read()
    atomic.SetUint32(&done, 1)
  }()
  Eventually(func() uint32 { return atomic.ReadUint32(&done) }).ShouldNot(BeZero())
  // Assertions
})
```

You still have the issue with the non-terminating go routine in case 2, right? If you want to get rid of async tests, could we think about making case 2 simpler (require less code)?

onsi:
Yes you do still have a non-terminating go routine in case 2.  But imagine a more complex test with more going on than a simple `Read()` in the `It` - by explicitly launching a goroutine around *just* the read you constrain which bit of the `It` is timing out and have greater control over what leaks in a non-terminating goroutine.

In any event, I agree that case 2 is ugly.  I'd probably use a channel:

```go
It("works", func() {
  var n int
  var err error
  done := make(chan struct{})
  go func() {
    defer GinkgoRecover()
    n, err = r.Read()
    close(done)
  }()
  Eventually(done).Should(BeClosed())
  // Assertions
})
```

but that's still a *lot* of code.

My instinct here is to create a helper wrapper that takes an `io.Reader` and a timeout `time.Duration` and returns an `io.Reader` that errors if the underlying reader doesn't return within the timeout.

I actually think something like that would be useful in Gomega.  In fact, stepping back, I think you've identified a missing Gomega feature and one that won't be too hard to add.  Are you familiar with Gomega's `gbytes` [package](http://onsi.github.io/gomega/#gbytes-testing-streaming-buffers)?

I think I can extend `gbytes.Buffer` so that it can wrap an `io.Reader` with an explicit timeout.  Then any call to the `buffer`'s `Read()` would return a known error in case of a timeout.  What this also lets you do is use `gbytes.Say` to make ordered assertions against the buffer.

Putting it all together, the [quic-go changes](https://github.com/lucas-clemente/quic-go/pull/724/files) that are driving this PR would end up going from:

```go
It("makes new streams", func(done Done) {
    sess.handleStreamFrame(&frames.StreamFrame{
        StreamID: 5,
        Data:     []byte{0xde, 0xca, 0xfb, 0xad},
    })
    p := make([]byte, 4)
    str, err := sess.streamsMap.GetOrOpenStream(5)
    Expect(err).ToNot(HaveOccurred())
    Expect(str).ToNot(BeNil())
    _, err = str.Read(p)
    Expect(err).ToNot(HaveOccurred())
    Expect(p).To(Equal([]byte{0xde, 0xca, 0xfb, 0xad}))
    close(done)
})
```

to something that looks more like:

```go
It("makes new streams", func() {
    sess.handleStreamFrame(&frames.StreamFrame{
        StreamID: 5,
        Data:     []byte{0xde, 0xca, 0xfb, 0xad},
    })
    str, err := sess.streamsMap.GetOrOpenStream(5)
    Expect(err).ToNot(HaveOccurred())
    Expect(str).ToNot(BeNil())
    Eventually(gbytes.BufferWithTimeout(str, time.Second)).Should(gbytes.Say([]byte{0xde, 0xca, 0xfb, 0xad}))
})
```

What do you think?

onsi:
That was easier than I thought it would be.  There's no need to wrap the `Reader` in a timeout.  With the latest Gomega you should be able to change your tests to:

```
It("makes new streams", func() {
    sess.handleStreamFrame(&frames.StreamFrame{
        StreamID: 5,
        Data:     []byte{0xde, 0xca, 0xfb, 0xad},
    })
    str, err := sess.streamsMap.GetOrOpenStream(5)
    Expect(err).ToNot(HaveOccurred())
    Expect(str).ToNot(BeNil())
    Eventually(gbytes.BufferReader(str).Contents).Should(Equal([]byte{0xde, 0xca, 0xfb, 0xad}))
})
```

I believe using `buffer.Contents` along with `Equal` gets you closer to your original intent (i.e. exactly _this_ set of bytes should be written down the wire).

lucas-clemente:
That looks like a nice addition to gbytes, thanks for the fix!

We have a similar case for io.Writer and io.Closer, where I guess gbytes won't really work. But you are right, there should be a way of building our own small helper that just takes a timeout, a Writer and (maybe) a matcher to do the assertion. Let me know if you come up with a better idea.

onsi:
Seems like something that should go in Gomega actually.  I'm going to add a few wrappers to `gbytes`:

`TimeoutCloser`
`TimeoutReader`
`TimeoutWriter`

which will all return `ErrTimeout` if the underlying `Closer`/`Reader`/`Writer` does not return in time.

onsi:
Done - take a look and lemme know if these end up being helpful to you!

lucas-clemente:
Looks like they are: https://github.com/lucas-clemente/quic-go/pull/742

Thanks! 👍

onsi:
👍 great!

Issue #205, opened 2016-01-04
Cover profile for focus tests


jasonkeene:
When Describe, It, etc are focused `ginkgo -cover` does not output a coverprofile at all. It would be nice if, when you focus tests, you get coverage statistics for just the tests that run. That would allow developers to easily pinpoint what code is being exercised by what tests. If this feature is not desired then I'd prefer some sort of message in the output letting you know that -cover is not supported while tests are focus.

onsi:
Sorry for the delay.  I've looked into this and it is, in fact, an issue - but I don't think it's fixable :/

Ginkgo tests exit with a "special" status code (197) if the user applies a programmatic focus (i.e. `FDescribe`, `FIt`, etc...).  The intent with this non-zero exit code is to ensure that CI systems will fail the test if there is a programmatic focus - it is almost always a mistake to commit a focused test!

Computing the coverage statistics is completely handled by `go test` which `ginkgo` calls out to.  Apparently `go test` does not compute coverage stats if the test process exits non-zero (or, perhaps, if it exits prematurely as Ginkgo must in order to set the exit code).  As far as I can tell, there's no way to exit 197 *and* get `go test` to emit coverage stats.

If, however, you use the `ginkgo --focus=<REGEXP>` flag Ginkgo does *not* exit with the special status code.  The idea here is that you've explicitly configured CI to only run a subset of tests.  In this case, `ginkgo --cover` *does* work.

robdimsdale:
@jasonkeene given the above comment, this is an issue that we can't fix in ginkgo, so I'm going to close this issue. Please feel free to open a new issue if you have any further questions.

jasonkeene:
@robdimsdale Yeah, I was thinking about closing this a little while ago as well. I've been using --focus and a few other work arounds to get the cover profiles I need. If these work arounds don't work in the future I'll reopen. Thanks!

joefitzgerald:
@onsi Sorry to resurrect this from the dead, but perhaps if there were a flag we could pass to suppress the `197` exit code, this could be used by integrations with editors to show coverage for focused blocks?

Obviously the default behavior should continue to be a non-zero exit code, but it feels like this would be a nice compromise to allow for both modes of use?

Are you open to a PR for this?

onsi:
[...]

PR #367, opened 2017-08-10
Allow coverage to be displayed for focused specs

joefitzgerald:
In #205, there was lamentation! Show coverage for focused specs, we could not 😢. With this PR, it is possible to opt-in to allow coverage to be shown for focused specs.

The default behavior (failing the suite if any focusing is present, to prevent it from getting through CI) is preserved. A savvy editor integrator can transparently set the `GINKGO_EDITOR_INTEGRATION` environment variable to some non-empty value (e.g. `true`) to suppress the `197` exit code, and allow a coverage file to be generated (and thus allow coverage to be displayed in the editor).

### DIFF
--- a/ginkgo/run_command.go
+++ b/ginkgo/run_command.go
@@ -5,14 +5,16 @@ import (
 	"fmt"
 	"math/rand"
 	"os"
+	"strings"
 	"time"
+
+	"io/ioutil"
+	"path/filepath"
 
 	"github.com/onsi/ginkgo/config"
 	"github.com/onsi/ginkgo/ginkgo/interrupthandler"
 	"github.com/onsi/ginkgo/ginkgo/testrunner"
 	"github.com/onsi/ginkgo/types"
-	"io/ioutil"
-	"path/filepath"
 )
 
 func BuildRunCommand() *Command {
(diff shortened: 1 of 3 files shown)


ginkgo/run_command.go
```diff
@@ -5,14 +5,16 @@ import (
 	"fmt"
 	"math/rand"
 	"os"
+	"strings"
 	"time"
 
+	"io/ioutil"
```
joefitzgerald: `goimports` doing it's thing. I'm happy to revert these lines if you'd prefer...

jasonkeene:
I'd ask for the environment variable to be slightly more generic. Perhaps just EDITOR_INTEGRATION=true This will make it easier to get into vim-go as I don't think they want to special case ginkgo.

joefitzgerald:
This is a very ginkgo specific thing. I would imagine a non-specific thing like `EDITOR_INTEGRATION` would be harder and not easier to gain acceptance? I actually thought I might need to be even more specific (e.g. `GINKGO_FOCUSED_COVERAGE`.

ginkgo/run_command.go
onsi: no problem :)

onsi:
i don't feel too strongly about what to call the env variable - perhaps get input from the vim-go maintainers?

@joefitzgerald can you add an integration test under `/integration` that asserts the exit code is zero for a suite with a focused test when the env variable is passed through?  thanks!

joefitzgerald:
@onsi Integration test added.

onsi:
Sweet thanks!

PR #372, opened 2017-08-26
Add test for MVC frameworks such as BeeGo or Revel

AlexeySoshin:
In those cases test files reside in different folder/package than the
actual code covered

### DIFF
--- a/integration/_fixtures/mvc_coverage_fixtures/controllers/controllers.go
+++ b/integration/_fixtures/mvc_coverage_fixtures/controllers/controllers.go
@@ -1,0 +1,31 @@
+package controllers
+
+import (
+	"encoding/json"
+	"strconv"
+)
+
+type MvcController struct{}
+
+func (m *MvcController) Get() (string, error) {
+
+	someMap := map[int]int{}
+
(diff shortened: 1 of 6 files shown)


williammartin:
Hey @AlexeySoshin I'm a little confused by this PR and feel I'm missing context having just started dealing with Ginkgo/Gomega. Can you explain a little more what these tests are adding here in terms of behavioural coverage, or how the current tests aren't sufficient? Happy to review once I understand this. Thanks!

AlexeySoshin:
Would be glad to.
What those test add is validation that frameworks such as [Beego](https://beego.me/) or [Revel](https://revel.github.io/) are properly covered by Ginkgo.
What happens with those frameworks is that they enforce a structure on your project such as:
- controllers
- models
- tests

So, the tests are outside of the package, and they also expected to test at least two other packages.

lflux:
@AlexeySoshin It seems this is already covered in https://github.com/onsi/ginkgo/blob/master/integration/coverage_test.go#L37-L41 - although it's not very explicitly called in the `By`.

AlexeySoshin:
@lflux Agreed. I'll close this PR then.

PR #373, opened 2017-08-26
Document -coverprofile and -outputdir flags

AlexeySoshin:


AlexeySoshin:
@onsi Could we merge this? :)

Issue #377, opened 2017-09-15
expose ability to alias It, Describe, etc like gomega allows with by changing skip


gaffo:
I have some wrapper functions It and Describe and I'd like them to show the correct context... but I can't access the Skip variable.

gaffo:
ping... per the contributing guide I wanted to discuss how to do this. I was considering exposing DescribeSkip, ItSkip, etc... but that seems somewhat clunky.

gaffo:
I think the more correct way would be to expose code point but that's more invasive

pmzajaczkowski:
I have the same issue. I am writing some wrappers around It, Describe, etc. so my tests look better and to have standardized output in tests failure raport, but the links to the code location then points to my functions (wrappers) and not to tests - becuase of fixed skip value inside for example It(...) function. I'd like to have some functions like CustomIt(..., skip = 3) or to have access to the globalSuite so I can write similiar code as is in the body of It(...) but with different skip value.

onsi:
I'm working through the backlog of old Ginkgo issues - apologies as this issue is probably stale now.

I'm going to address this as part of the [v2](#711) work - I'm going to make it possible to decorate tests with an `Offset(N)` to shift the line numbers up the call stack.

I'm going to close this for now, but feel free to reopen it if you'd like.

onsi:
A beta for Ginkgo V2 that fixes this issue was just released.   [Please try out the beta](https://github.com/onsi/ginkgo/issues/809)!

PR #382, opened 2017-10-12
Exposing skip for evaluation by onsi

gaffo:
Opening for dicussion for fixing #377

### DIFF
--- a/ginkgo_dsl.go
+++ b/ginkgo_dsl.go
@@ -283,6 +283,12 @@ func Describe(text string, body func()) bool {
 	return true
 }
 
+// You can wrap the method in your own meta framework and still get valid errors
+func DescribeSkip(text string, skip int, body func()) bool {
+	globalSuite.PushContainerNode(text, body, types.FlagTypeNone, codelocation.New(skip + 1))
+	return true
+}
+
 //You can focus the tests within a describe block using FDescribe
 func FDescribe(text string, body func()) bool {
 	globalSuite.PushContainerNode(text, body, types.FlagTypeFocused, codelocation.New(1))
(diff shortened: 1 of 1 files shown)


onsi:
@gaffo this seems like a reasonable enough approach to me.  Yeah, it's clunky, but it's also explicit and relatively simple so I'm in favor.  Wanna add some tests and some documentation (under the gh-pages branch)?

gaffo:
Which test file should I put em in?

onsi:
i’d suggest doing it as an integration test - you should be able to follow the patterns in the /integration folder.

gaffo:
Any specific file I should dump it in in integration, next to something else? do I need to check the error text is correct or just make sure it compiles and runs correctly?

onsi:
Actually - now that I think of it `DescribeSkip` might be confusing since there is a `Skip` already.  Could we mirror [Gomega](https://github.com/onsi/gomega/blob/master/gomega_dsl.go#L137) and use `DescribeWithOffset` etc.?

I think you could then create a new `integration/offset_test.go`.  And yeah I think we'd want to make sure the error text is correct to catch regressions.  An easy way to do that is make a new fixture, run the test, grab the line number, then just hard code that in the test.  You _should_ be able to pattern match and copy/paste your way to victory with the integration tests but let me know if I can help.

gaffo:
Okay thanks that's the advice I was looking for!

williammartin:
@gaffo Looks like maybe this has been languishing, do you think you'll get a chance to implement Onsi's suggestions?

gaffo:
Yes... Just waiting for time.

pmzajaczkowski:
It looks like all of the conditions are met to merge-in this pull request. Could you accept it? :)

williammartin:
@pmzajaczkowski Changes have been requested earlier in the comment chain. If you'd like to tackle those and open a separate PR I'd be happy to accept that.

PR #383, opened 2017-10-13
Added Duration on GinkgoTestDescription

eloycoto:
Hello, 

This PR adds the support to get the test Duration on the CurrentGinkgoTestDescription(). The main reason is when a test fails you know the current duration of the test.

In my particular case, when a test fails, logs can be retrieved since the test starts. So I don't need to add more verbose in the current test output. 

Regards

### DIFF
--- a/ginkgo_dsl.go
+++ b/ginkgo_dsl.go
@@ -149,7 +149,8 @@ type GinkgoTestDescription struct {
 	FileName   string
 	LineNumber int
 
-	Failed bool
+	Failed   bool
+	Duration time.Duration
 }
 
 //CurrentGinkgoTestDescripton returns information about the current running test.
@@ -169,6 +170,7 @@ func CurrentGinkgoTestDescription() GinkgoTestDescription {
 		FileName:       subjectCodeLocation.FileName,
 		LineNumber:     subjectCodeLocation.LineNumber,
 		Failed:         summary.HasFailureState(),
+		Duration:       summary.RunTime,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


eloycoto:
@onsi  sorry for bothering you, any issue with this PR? Regards

williammartin:
Hey @eloycoto just making sure I understand the PR in context. Are you doing something like `CurrentGinkgoTestDescription` in an `AfterEach` block, so you need the Duration to be consistently updating throughout the run?

eloycoto:
Hi @williammartin 

Yes! I need the current duration of the test in AfterEach.

williammartin:
Alright @eloycoto I'm on board with this functionality but I think there's a few things that need to be resolved around this PR.

Functionally, I think it is a problem that the runTime here will continue to increase after the test has finished when fetched via `Summary()`. See this test that could live in `spec_test.go`:

```
It("should have a runtime which remains consistent after spec run", func() {
	totalRunTime := summary.RunTime
	Ω(totalRunTime).Should(BeNumerically(">=", 10*time.Millisecond))

	Consistently(func() time.Duration { return spec.Summary("suite id").RunTime }).Should(Equal(totalRunTime))
})
```

I think this can be resolved by something as simple as:

```
runTime := spec.runTime
if runTime == 0 {
	runTime = time.Since(spec.startTime)
}
```

and then setting 		

```
RunTime:      runTime,
```

on the spec summary.

The other thing I think required here is a test for the behaviour you actually want, which is that:

`It increases the runTime throughout the test execution` 

Because as far as I can see the only test that exists is post-execution. I think it would be fair for this test to live on the GinkgoCurrentTestDescription level, or the Summary level.

WDYT about these changes?

Thanks!

eloycoto:
I'll work on this in the following days!

eloycoto:
Hi @williammartin this should be ok to review ;-) 

Sorry, It has been a busy week!

williammartin:
Alright, hopefully will get round to this shortly now Christmas is over :)

eloycoto:
@williammartin Fixed ;-)

williammartin:
Cool. Thanks.

Issue #348, opened 2017-05-01
Add a When() synonym for Context()


decibel:
I find myself writing a lot of `Context()`s with a description in the form "when <something>", ... It(). Enough so that I'm wondering if it'd be worth adding a When() to the DSL, something along the lines of:

```func When(description string, ...) ... {
    return Context("when " + description, ...)
}```

That way if you write:

```When("something happens", func() {
...
})```

you'd get a "when" in the text output as well.

onsi:
Makes sense and should be relatively easy.  Will try to get to this soon.

williammartin:
This was done in https://github.com/onsi/ginkgo/pull/386. Hope it works for you.

decibel:
Works great, thanks for your efforts!

PR #386, opened 2017-10-30
Add a When() synonym for Context()

akshaymankar:
Solves #348

### DIFF
--- a/ginkgo_dsl.go
+++ b/ginkgo_dsl.go
@@ -275,9 +275,9 @@ func GinkgoRecover() {
 //Describe blocks allow you to organize your specs.  A Describe block can contain any number of
 //BeforeEach, AfterEach, JustBeforeEach, It, and Measurement blocks.
 //
-//In addition you can nest Describe and Context blocks.  Describe and Context blocks are functionally
+//In addition you can nest Describe, Context and When blocks.  Describe, Context and When blocks are functionally
 //equivalent.  The difference is purely semantic -- you typical Describe the behavior of an object
-//or method and, within that Describe, outline a number of Contexts.
+//or method and, within that Describe, outline a number of Contexts and Whens.
 func Describe(text string, body func()) bool {
 	globalSuite.PushContainerNode(text, body, types.FlagTypeNone, codelocation.New(1))
 	return true
@@ -304,9 +304,9 @@ func XDescribe(text string, body func()) bool {
 //Context blocks allow you to organize your specs.  A Context block can contain any number of
 //BeforeEach, AfterEach, JustBeforeEach, It, and Measurement blocks.
 //
-//In addition you can nest Describe and Context blocks.  Describe and Context blocks are functionally
+//In addition you can nest Describe, Context and When blocks.  Describe and Context blocks are functionally
 //equivalent.  The difference is purely semantic -- you typical Describe the behavior of an object
-//or method and, within that Describe, outline a number of Contexts.
+//or method and, within that Describe, outline a number of Contexts and Whens.
 func Context(text string, body func()) bool {
 	globalSuite.PushContainerNode(text, body, types.FlagTypeNone, codelocation.New(1))
 	return true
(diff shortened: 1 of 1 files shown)


williammartin:
Thanks for the PR @akshaymankar I'm gonna tag this as `needs-review` because there's a bunch of changes in here, but definitely on board with the premise!

lflux:
Looks good, would you be open to rebasing off master and repushing so we can get coverage on 1.9 now that our Travis configuration runs against it? I ran your branch locally on 1.9.1 and it failed, but if I rebased off master tests passed.

lflux:
I like this functionality! In my mind, `When` is a special case of `Context` and we should treat it as such in our tests. I don't feel that it warrants changing all the existing `Context`s that are in the tests. Are you open to adding tests to add coverage for `When`, but backing out the `Context("When...` -> `When(` changes?

adamcohen:
Good to see this PR getting some traction.  I've been looking forward to seeing it get merged since I use `When` quite frequently in my tests.

akshaymankar:
@lflux Thanks for the review. I will rebase the branch in my flex-hour or over the weekend.

I couldn't find a place to add tests for dsl code. If you could point me to it, I'll add tests for the functions related to `When`.

I am not sure why should `Context("When..` to`When(...` be backed out, could you please clarify?

williammartin:
I don't _really_ want this merged because it opens the door to DSL bloat for all the conjunctions e.g. `And(`,  `But(`, `Because(`, `AndWhen(` (cause nesting `When(` in `When(` is weird), etc. Less problematic, it also means that during a refactoring stage you could pretty easily change textual description but forget to change correlated method call, leaving confusing test output.

Pragmatically though, I understand that people are interested in this and there is some precedent for `It(` and `Specify(` in https://github.com/onsi/ginkgo/commit/e986a52d59247313ed8ef4870d697e515578644 so it's probably fine to do so but I'd like a few things changed.

Seems like instead of doing:

```
globalSuite.PushContainerNode("when "+text, body, types.FlagTypeFocused, codelocation.New(1))
return true
```

we could do:

```
return Context(.....
```

In the same way specify does, just to DRY this out a little.

Due to my reasons at the top, I'd also prefer not to see all the Contexts changed to Whens, and rather have it a special case.

Finally, definitely want some docs to be PRed along with this to the `gh-pages` branch.

TL;DR:

Changes I'd like:

1. DRY When by following the Specify example
2. Test `When(` as a special case
3. PR docs.

But I'm happy to chat about any/all of these! 

Thanks.

akshaymankar:
I am not so sure about the DSL bloat problem. Right now it is not such a big problem, but if it does start bloating maybe that can be a separate project.

My views/comments on the requested changes:
> 1. DRY When( by following the Specify( example

I already have a problem with the Specify works. Whenever I have a failure, ginkgo tells me the failure is in ginkgo's code where Specify is defined. As somebody who has seen his tests red, I wouldn't want ginkgo's line numbers in my report.

> 2. Test When( as a special case

I agree with the arguments given. I will revert changes to the tests and add another test.

> 3. PR docs

I will write it up and send another PR.

williammartin:
@akshaymankar I didn't realise `Specify(` printed a stack trace like that (yuck!). I'm not able to reproduce that:

```
Specify("doesn't print the right stack trace", func() {
	Fail("fail the test")
})
```

->

```
Users/pivotal/go/src/github.com/williammartin/ginkgoleaktest/ginkgoleaktest_test.go:8
  doesn't print the right stack trace [It]
  /Users/pivotal/go/src/github.com/onsi/ginkgo/ginkgo_dsl.go:394

  fail the test

  /Users/pivotal/go/src/github.com/williammartin/ginkgoleaktest/ginkgoleaktest_test.go:11
```

Totally agree with you if that is what happens though. Thanks.

akshaymankar:
@williammartin Third line in the output that you posted:

```
/Users/pivotal/go/src/github.com/onsi/ginkgo/ginkgo_dsl.go:394
```

It usually prints the line where `It` is called, but for specify it is always `ginkgo_dsl.go:394`

williammartin:
Doh! Thanks for keeping me right! I created https://github.com/onsi/ginkgo/issues/414 if you want to continue that discussion elsewhere. Otherwise I'm good with this if we fix up the special case `When` test. Thanks.

williammartin:
Didn't see that you'd updated this PR, sorry! Happy to merge now.

Two outstanding things we can follow up on after is updating the docs, and writing a test, maybe within the `progress_fixture` that makes use of a `When` block. Also some comments in `internal/suite/suite.go` and `README.md` about nestable DSL, but let's not block this any more.

Thanks

PR #389, opened 2017-11-07
Added Stdin support to TestRunner

eloycoto:
Hello! 

A minimal change that allows using stdin on ginkgo test runner. 

The main reason to allow Stdin is to be able to wait for test on a custom fail function, that allows developers to debug in some particular circumstances. 

Regards

### DIFF
--- a/ginkgo/testrunner/test_runner.go
+++ b/ginkgo/testrunner/test_runner.go
@@ -431,6 +431,7 @@ func (t *TestRunner) cmd(ginkgoArgs []string, stream io.Writer, node int) *exec.
 	cmd.Dir = t.Suite.Path
 	cmd.Stderr = io.MultiWriter(stream, t.stderr)
 	cmd.Stdout = stream
+	cmd.Stdin = os.Stdin
 
 	return cmd
 }


williammartin:
Hey @eloycoto can you give me some background on this, and how you are using it more concretely? Are there some tests we can write for this behaviour?

eloycoto:
Hey @williammartin, 

On Cilium project, we're using ginkgo to make e2e testing. Sometimes, for devs, is easy when it fails to wait and check what the situation on the servers is. 

If you have the stdin enabled you can ask for a key to continue, and meanwhile you can debug manually in another terminal. 

At the moment we are using a custom flag and this code, that it's not pretty: 

```
if config.CiliumTestConfig.HoldEnvironment {

      
		test := ginkgo.CurrentGinkgoTestDescription()
		pid := syscall.Getpid()

		fmt.Fprintf(os.Stdout, "\n---\n%s", test.FullTestText)
		fmt.Fprintf(os.Stdout, "\nat %s:%d", test.FileName, test.LineNumber)
		fmt.Fprintf(os.Stdout, "\n\n%s", description)
		fmt.Fprintf(os.Stdout, "\n\nPausing test for debug, use vagrant to access test setup.")
		fmt.Fprintf(os.Stdout, "\nRun \"kill -SIGCONT %d\" to continue.\n", pid)
		syscall.Kill(pid, syscall.SIGSTOP)
	}
```

Around write the test, not sure in where.. but if you point me in the right direction I'll do. 

Regards

williammartin:
This makes sense, I just wonder if there is a good integration test we could write for it.

eloycoto:
Hi @williammartin 

could you help me a bit with this? No idea how/where to make the test :-(

Regards

williammartin:
Well, to put the question another way, if I merged this, what code would you write in your tests to make use of it? Once we understand that, we can make sure to write a regression test to make sure we don't break you in future.

eloycoto:
Hi @williammartin 

If is valid for you if I changed the ginkgo line in travis to something like: 
```
while true; do;  echo "Y"; done | $HOME/gopath/bin/ginkgo -r --randomizeAllSpecs --randomizeSuites --race --trace  && go vet
```

If not, I do not found any other way to test the stdin support

PR #400, opened 2017-12-06
Remove unused variable

alex-slynko:
Are there any missing tests?


williammartin:
Not sure about the missing tests. Looks like the `pendingFlag` is used to determine that specs marked pending are not run by the runner. I wonder whether there is expected behaviour tests for focused or if they exist elsewhere. Maybe @onsi knows.

PR #403, opened 2017-12-06
Fix go vet issues

alex-slynko:


internal/leafnodes/shared_runner_test.go
```diff
@@ -180,7 +180,6 @@ func AsynchronousSharedRunnerBehaviors(build func(body interface{}, timeout time
 					failer.Fail("bam", innerCodeLocation)
 					time.Sleep(20 * time.Millisecond)
 					panic("doesn't matter")
-					close(done)
```
williammartin: I think this should be in a defer pre-panic. Same for below.

alex-slynko:
@williammartin Ready!

williammartin:
LGTM thanks!

PR #404, opened 2017-12-06
Build with go 1.9 in CI

lflux:
Gomega builds with 1.6-1.9, so let's be consistent. Also, drop support
for 1.5 since it's really old


williammartin:
LGTM. 1.5 is old news!

Issue #406, opened 2017-12-08
Are the version numbers in Ginkgo docs REALLY accurate?


williammartin:
There's a few mentions of supported versions of golang and versions in Ubuntu apt repos in the docs and I'm not sure they are accurate.

williammartin:
We should probably update version numbers to reflect what we run in CI, outside of that it's unsupported.

PR #409, opened 2017-12-13
Update README with Go 1.6+, Golang -> Go

lflux:
Address parts of #406 for the README
- We support Go 1.6+ in CI, in line with what the Go project supprots
- Point to official Go installation instructions
- Replace Golang with Go


williammartin:
LGTM, maybe not "need" 1.6 but suggest due to what we support/test. Also maybe if you push the build will run again and we can see that failure is a flake heh

PR #410, opened 2017-12-13
Reflect latest Go versions

lflux:
- Our CI only supports 1.6+ currently, so let's reflect that in the docs
- Replace Golang with Go
- Remove ubuntu-specific callout and replace with generic installation instructions

Address #406

### DIFF
--- a/index.md
+++ b/index.md
@@ -2,7 +2,7 @@
 layout: default
 title: Ginkgo
 ---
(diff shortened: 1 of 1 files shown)


williammartin:
LGTM but maybe instead of "compatible" we should say "supported"? WDYT?

lflux:
Reworded to `tested against`

Issue #394, opened 2017-12-01
Flakey / leaky integration tests


hoegaarden:
When we run the tests as described in the contribution guide (`ginkgo -r -p`), we see that the test for the coverage integration tests are very flakey.
We think that is because of race condition in `integration/coverage_test.go` when tests run in parallel -- in `AfterEach` a fixtures directory is being removed while other tests, depending on that, are potentially running.

The error we see quite often is (similar to):
```
• Failure [3.115 seconds]
Coverage Specs [It] runs coverage analysis in series and in parallel
/Users/pivotal/go/src/github.com/onsi/ginkgo/integration/coverage_test.go:18

  Expected error:
      <*exec.ExitError | 0xc4200ec880>: {
          ProcessState: {
              pid: 86028,
              status: 256,
              rusage: {
                  Utime: {
                      Sec: 0,
                      Usec: 25735,
                      Pad_cgo_0: [128, 255, 255, 255],
                  },
                  Stime: {
                      Sec: 0,
                      Usec: 36515,
                      Pad_cgo_0: [0, 0, 0, 0],
                  },
                  Maxrss: 9101312,
                  Ixrss: 0,
                  Idrss: 0,
                  Isrss: 0,
                  Minflt: 11850,
                  Majflt: 3,
                  Nswap: 0,
                  Inblock: 0,
                  Oublock: 0,
                  Msgsnd: 0,
                  Msgrcv: 0,
                  Nsignals: 3,
                  Nvcsw: 374,
                  Nivcsw: 735,
              },
          },
          Stderr: nil,
      }
      exit status 1
  not to have occurred

  /Users/pivotal/go/src/github.com/onsi/ginkgo/integration/coverage_test.go:32
------------------------------
```

The revision we have seen this in is 7a241e9c6c7fb89adf3965bd763b7369b88a51eb, we believe that this problem already exists longer, but we didn't triage it.

How to reproduce:
- run `ginkgo -r --randomizeAllSpecs --randomizeSuites --race --trace -p -untilItFails -focus 'Coverage' integration/` in the root of the project
- See that the tests **fail** pretty fast (for us consistently on 1st or 2nd run currently)
- run `ginkgo -r --randomizeAllSpecs --randomizeSuites --race --trace -p -untilItFails integration/` (without focusing on the coverage tests)
- See that the tests **fail** after a couple of runs, not as consistent as the one above
[...]

PR #423, opened 2018-01-24
Fix race condition in coverage tests

alamages:
Tests now generate coverprofile files into different paths so now they don't
interfere with other and can be run successfully in parallel. Also since
the files now have different names, some extra cleanup was required.

Related issue: #394

### DIFF
--- a/integration/coverage_test.go
+++ b/integration/coverage_test.go
@@ -1,115 +1,124 @@
 package integration_test
 
 import (
-	"os"
 	"os/exec"
 
 	"fmt"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gbytes"
 	"github.com/onsi/gomega/gexec"
 )
 
 var _ = Describe("Coverage Specs", func() {
-	AfterEach(func() {
-		os.RemoveAll("./_fixtures/coverage_fixture/coverage_fixture.coverprofile")
-	})
+	Context("when it runs coverage analysis in series and in parallel", func() {
+		AfterEach(func() {
+			removeSuccessfully("./_fixtures/coverage_fixture/coverage_fixture.coverprofile")
+		})
+		It("works", func() {
+			session := startGinkgo("./_fixtures/coverage_fixture", "-cover")
+			Eventually(session).Should(gexec.Exit(0))
 
-	It("runs coverage analysis in series and in parallel", func() {
-		session := startGinkgo("./_fixtures/coverage_fixture", "-cover")
-		Eventually(session).Should(gexec.Exit(0))
-		output := session.Out.Contents()
-		Ω(string(output)).Should(ContainSubstring("coverage: 80.0% of statements"))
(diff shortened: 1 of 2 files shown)


williammartin:
Thanks for looking at this, will try to review today.

integration/coverage_test.go
```diff
@@ -59,23 +58,24 @@ var _ = Describe("Coverage Specs", func() {
...
-		session := startGinkgo("./_fixtures/combined_coverage_fixture", "-r", "-cover", "-coverprofile=coverage.txt")
+		session := startGinkgo("./_fixtures/combined_coverage_fixture", "-r", "-cover", "-coverprofile=coverage1.txt")
 
 		Eventually(session).Should(gexec.Exit(0))
 
 		packages := []string{"first_package", "second_package"}
 
 		for _, p := range packages {
-			coverFile := fmt.Sprintf("./_fixtures/combined_coverage_fixture/%s/coverage.txt", p)
+			coverFile := fmt.Sprintf("./_fixtures/combined_coverage_fixture/%s/coverage1.txt", p)
```
williammartin: I wonder whether we can come up with more self-explaining names for these files than 1,2,3 etc. e.g. `coverage-recursive.txt`

I'm also not super clear on why these all have different names rather than different output dirs but I'm definitely missing something.

williammartin:
Bump @alamages any thoughts on the review?

integration/integration_suite_test.go
```diff
@@ -89,3 +89,8 @@ func startGinkgo(dir string, args ...string) *gexec.Session {
 	Ω(err).ShouldNot(HaveOccurred())
 	return session
 }
+
+func removeSuccessfully(path string) {
```
williammartin: Can we just use `Expect(os.RemoveAll(path)).To(Succeed())` ? Do we need this extra indirection to another method as well, or is it cause you don't want the assertion in the test block itself?

alamages: Yes, the initial idea was not to have the assertion within the block itself. The PR is still WIP since we started refactoring to make sure all generated coverage files are always removed.

alamages:
@williammartin you can take a look at this PR again now

We specify different output coverage files instead of different output directories because:
- coverage filenames where already being provided in the tests, we just had to change their names so they won't conflict
- specifying different output directory for coverage in recursive mode would only affect the combined coverage file. The intermediates package coverage files would still be generated in the default path.

Let me know if you want any further changes on this, thanks!

alamages:
Once this PR: https://github.com/onsi/ginkgo/pull/446 is merged I will have to merge with master and these changes will work with go 1.10 as well.

williammartin:
@alamages gogogo

alamages:
@williammartin  done :D

williammartin:
Can you fix up the commits as well please? I'd rather not have a git tree containing a bunch of `wip` commits.

alamages:
@williammartin it should be ok now

williammartin:
LGTM thanks

Issue #436, opened 2018-03-02
Ginkgo dryRun outputs "slow tests"


nodo:
When running `ginkgo` with the `-dryRun` flag, it wrongly warns the user about slow tests.

For instance using this test:

```go
var _ = Describe("Playground", func() {
	It("succeeds", func() {
		Expect(1).To(Equal(1))
	})
})
```

It outputs the following:

```bash
$> ginkgo -dryRun .
Running Suite: Playground Suite
===============================
Random Seed: 1519999176
Will run 1 of 1 specs

• [SLOW TEST:9223372036.855 seconds]
Playground
/Users/pivotal/go/src/github.com/nodo/playground/playground/playground_test.go:8
  succeeds
  /Users/pivotal/go/src/github.com/nodo/playground/playground/playground_test.go:9
------------------------------

Ran 1 of 1 Specs in 0.000 seconds
SUCCESS! -- 0 Passed | 0 Failed | 0 Pending | 0 Skipped PASS

Ginkgo ran 1 suite in 612.596828ms
Test Suite Passed
```

Note that the test took 612ms to run, but the warns says it took 9223372036.855 seconds. We would expect to not see the warning.

with @kirederik

PR #438, opened 2018-03-07
Do not set the run time of a spec when the dryRun flag is used

MirahImage:
Fixes https://github.com/onsi/ginkgo/issues/436


kirederik:
🎉

Issue #421, opened 2018-01-19
Gingko needs a releasing guide


williammartin:
See Gomega: https://github.com/onsi/gomega/blob/master/RELEASING.md

PR #441, opened 2018-03-13
Add a `RELEASING.md` guide for creating a release

blgm:
- resolves https://github.com/onsi/ginkgo/issues/421
- based on https://github.com/onsi/gomega/blob/master/RELEASING.md
- only @onsi has released so far, so it would make sense for him to review this
- @williammartin, a releasing guide is a step in the right direction, but how would be be if there were a releasing script that did it all for you?  Let's discuss...


williammartin:
LGTM thanks. An automated process would be good, but I'm not sure we do this enough to require infrastructure, are you just thinking a script?

PR #355, opened 2017-05-29
Handle -coverprofile flag

AlexeySoshin:
Fixes #346

PR #446, opened 2018-03-29
Fail fast when the outputdir of combined coverage profile does not exist

nodo:
This is a workaround for a change in behaviour between go <= 1.9 and go
1.10.

Prior to go 1.9, compiling a test would not fail if the specified
output dir does not exist. However, it does fail in go 1.10.

By failing when the the directory does not exist, we have a consistent
behaviour withing go versions.

See also https://github.com/golang/go/issues/24588

### DIFF
--- a/integration/coverage_test.go
+++ b/integration/coverage_test.go
@@ -112,17 +112,12 @@ var _ = Describe("Coverage Specs", func() {
 		os.RemoveAll("./_fixtures/combined_coverage_fixture/coverage.txt")
 	})
 
-	It("Creates directories in path if they don't exist", func() {
+	It("Fails with an error if output dir and coverprofile were set, but the output dir did not exist", func() {
 		session := startGinkgo("./_fixtures/combined_coverage_fixture", "-outputdir=./all/profiles/here", "-r", "-cover", "-coverprofile=coverage.txt")
 
-		defer os.RemoveAll("./_fixtures/combined_coverage_fixture/all")
-		defer os.RemoveAll("./_fixtures/combined_coverage_fixture/coverage.txt")
-
-		Eventually(session).Should(gexec.Exit(0))
-
-		_, err := os.Stat("./_fixtures/combined_coverage_fixture/all/profiles/here/coverage.txt")
-
-		Ω(err).ShouldNot(HaveOccurred())
+		Eventually(session).Should(gexec.Exit(1))
+		output := session.Out.Contents()
+		Ω(string(output)).Should(ContainSubstring("Unable to create combined profile, outputdir does not exist: ./all/profiles/here"))
 	})
 
 	It("Moves coverages if only output dir was set", func() {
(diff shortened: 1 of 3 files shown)


integration/coverage_test.go
```diff
@@ -112,17 +112,12 @@ var _ = Describe("Coverage Specs", func() {
 		os.RemoveAll("./_fixtures/combined_coverage_fixture/coverage.txt")
 	})
 
-	It("Creates directories in path if they don't exist", func() {
+	It("Fails with an error if cannot combine the suite coverage profiles because the output dir does not exist", func() {
```
blgm: Could we possible put this in a `Describe`?
```
Describe("when the output dir does not exist", func() {
  It("fails with an error", func() {
```
I don't think the text "cannot combine the suite coverage profiles" adds very much, so I would omit it.

nodo: The reason I specified "cannot combine the suite coverage profiles" is that by omitting the `-coverageprofile` flag, ginkgo will try to move the separate coverage profiles into the new directory without combining them.

As far as I understand, this is handled in another function (see [here](https://github.com/onsi/ginkgo/blob/master/ginkgo/run_command.go#L114), which has also different tests. 

How can we change the text to be clearer and explain that this test is specifically about `-coverageprofile` flag?

nodo:
Thanks for your review @blgm !

Adding an item to the changelog is definitely a good idea. I'll do it.

integration/coverage_test.go
blgm:
```
Describe("when combining coverage profiles", func() {
  Describe("when the output dir does not exist", func() {
    It("fails with an error", func() {
```
A very long `It()` is a smell, but I'm not sure this is any better.

williammartin: As an aside from the general discussion, I'd suggest using `Context(` in the case where the thing under test is undergoing different preconditions rather than `Describe(`. In fact we also have `When(` which is an alias for `Context(` and prefixes `when` to the description.

nodo: @blgm I have changed the description to be "Fails with an error if output dir and coverprofile were set, but the output dir did not exist", which follows the same structure to the `It` above.

In my opinion, adding a level of nesting such as `Context`, `When`, or `Describe` would make the tests less readable, but I agree that it was a very long `It` description.

What do you think about the message now?

blgm: @nodo, I think the message is clear, and definitely good enough.  I take your point regarding nesting.  I think it can lead to clearer code, but it would be out of step with the rest of the file, and consistency is more important.

nodo:
@blgm updated the changelog, do you think that's enough?

blgm:
Looks good to me

blgm:
I'm not considering this a breaking change because the `-coverprofile` flag has not yet been released, it was only added in https://github.com/onsi/ginkgo/pull/355.

PR #448, opened 2018-04-02
Added name field to jUnit reported testsuite

melnikk:
JUnit file format specification implies presense of "name" attribute. So, TeamCity parser substitutes "null" instead of Suite Name. PR fixes this issue.


williammartin:
LGTM thanks for the PR.

PR #451, opened 2018-04-04
Update fixtures to work using go1.10

nodo:
In go 1.10 the `-coverpkg` flag does not load new packages, but allow
users to provide a list of patterns (see
https://golang.org/doc/go1.10#test). For this reason, we need to import
the the package in the fixture.


nodo:
For context, this PR fixes a red test in go 1.10 (see [here](https://wings.pivotal.io/teams/ginkgo-gomega-contributors/pipelines/main/jobs/ginkgo/builds/2610#L5ab28826:211))

PR #456, opened 2018-04-12
Ignore packages whose tests are all ignored by go

idoru:
When a package contains only tests in go source code that are ignored (filenames being with `.` or `_`) I think ginkgo ought to avoid selecting this package for 2 reasons: 

1. the tests won't be compiled
2. if all other (non-test) `*.go` files in a packager are also ignored this leads to a failure: `can't load package: package X: no Go files in /path/to/X`


williammartin:
LGTM thanks.

PR #487, opened 2018-06-16
gofmt

whilei:
Just ran

	gofmt -w .

on the project root. That's all.

> https://blog.golang.org/go-fmt-your-code


blgm:
Looks good, thanks @whilei!

PR #501, opened 2018-07-13
🖋 Typo: excercise → exercise

cunnie:
We exercise caution when excising typos, for we are nothing if not careful.


nodo:
Thanks @cunnie 👌

Issue #234, opened 2016-03-14
Ginkgo panics when using the Table extension


robdimsdale:
see https://sunrise.ci.cf-app.com/pipelines/pivnet-resource/jobs/test/builds/135

```
[2] /tmp/ginkgo371214560/validator.test flag redefined: ginkgo.seed
[2] panic: /tmp/ginkgo371214560/validator.test flag redefined: ginkgo.seed
```

I see the same thing when locally when not running in parallel.

This is running golang 1.6 and I believe the behavior is not observed on golang 1.5 as I have used the `Table` extension in other projects on golang 1.5 without issue.

onsi:
very weird... i can't get this to reproduce locally with a minimal example.  maybe we can poke at it a bit tomorrow, @robdimsdale ?

robdimsdale:
I ran this with go 1.5.3 and it worked - https://sunrise.ci.cf-app.com/builds/855 is a one-off build demonstrating this behavior (you can't see the SHA but it's the same commit that fails in golang 1.6).

robdimsdale:
https://github.com/pivotal-cf-experimental/pivnet-resource/commit/95ee3d357f5adaa41f60f0f31394aa4bb24841f5 is the offending commit

robdimsdale:
Stepping away from the specifics of the test for that project and using a minimal example, I can reproduce locally this with the table-driven-testing example given at http://onsi.github.io/ginkgo/#table-driven-tests with the following set of files:

foo.go
```
package foo
```

foo_suite_test.go
```
package foo_test

import (
	. "github.com/onsi/ginkgo"
	. "github.com/onsi/gomega"

	"testing"
)

func TestFoo(t *testing.T) {
	RegisterFailHandler(Fail)
	RunSpecs(t, "Foo Suite")
}
```

foo_test.go
```
import (
    . "github.com/onsi/ginkgo/extensions/table"

    . "github.com/onsi/ginkgo"
    . "github.com/onsi/gomega"
)

var _ = Describe("Math", func() {
    DescribeTable("the > inequality",
        func(x int, y int, expected bool) {
            Expect(x > y).To(Equal(expected))
        },
        Entry("x > y", 1, 0, true),
        Entry("x == y", 0, 0, false),
        Entry("x < y", 0, 1, false),
    )
})
```

and I get the following output:
```
[...]

PR #502, opened 2018-07-14
Discussion point: ability to specify Flags prefix

carldunham:
This is a hacky way to do this, but should work. Open to other options.


nodo:
Thanks for you PR @carldunham, I really appreciate that you took the time to fix the docs for contributions and `go vet`. If you wish, I would be happy to merge those in a separate PR and keep this one for discussing the prefix problem.

Coming back to your PR, what is the actual problem you would like to solve? 

As far as I understand with this PR the compiled test file will have duplicate options with `ginkgo` and `ginkgo0` prefix (in case `init()` runs only once). I think this duplication is not desirable.  However, I would not change the flags prefix (at least by default), as I am fairly sure there are users of `ginkgo` that use `go test` directly.

carldunham:
@nodo Happy to split the PR. I'll actually submit that separately.

As for the extant issue, I ran into this basically because I was refusing to bend to Go's will with respect to project structure, mainly due to existing project constraints (multiple languages, other devs, existing conventions, etc). It also seems to be the same problem raised in #234.

I provided a toy project that also illustrates the issue in https://github.com/carldunham/go-dep-test.

It should be something that could be fixed in later versions of Go (1.11 module support, perhaps), but still could arise in cases where different versions of Ginkgo were required by different modules, say.

The suggestion in the PR is foolish, as you gently _didn't_ say 😸, but I wasn't able to do something cleaner in a short time. A more rational approach might just be to avoid setting the flags on subsequent calls to `init()`, but I was curious about the existence of `prefix` and wondered what I could do with it.

Thanks!

jianzhangbjz:
Still get the `flag redefined: ginkgo.seed` error even if use this fixed PR. Do you know how to solve it? Thanks!

```console
mac:extended-platform-tests jianzhang$ ./extended-platform-tests run all --dry-run
./extended-platform-tests flag redefined: ginkgo.seed
panic: ./extended-platform-tests flag redefined: ginkgo.seed

goroutine 1 [running]:
flag.(*FlagSet).Var(0xc000178540, 0x8757400, 0xd0c57a0, 0xc00364a042, 0xb, 0x7c2bd0e, 0x2a)
	/usr/local/Cellar/go/1.13.1/libexec/src/flag/flag.go:848 +0x4ae
flag.(*FlagSet).Int64Var(...)
	/usr/local/Cellar/go/1.13.1/libexec/src/flag/flag.go:673
github.com/openshift/origin/vendor/github.com/onsi/ginkgo/config.Flags(0xc000178540, 0x7b541f6, 0x6, 0x5719001)
	/Users/jianzhang/goproject/src/github.com/openshift/origin/vendor/github.com/onsi/ginkgo/config/config.go:68 +0x114
github.com/openshift/origin/vendor/github.com/onsi/ginkgo.init.0()
	/Users/jianzhang/goproject/src/github.com/openshift/origin/vendor/github.com/onsi/ginkgo/ginkgo_dsl.go:58 +0x10a
```

Issue #504, opened 2018-07-16
Ginkgo with Golang 1.10 attempts to overwrite system-installed versions of Golang


joestringer:
In my environment, I run Ubuntu 18.04 LTS with Golang 1.10 installed via Snap:

```
$ ls -l $GOROOT
lrwxrwxrwx 1 root root 4 Jun 22 10:56 /snap/go/current -> 2130
```

When I attempt to build my ginkgo tests, it fails due to this:

```
$ ginkgo build                                                                                                                                  
Compiling test...                                                                                                                                                                               
Failed to compile test:                                                                                                                                                                                                                 
                                                                                                                                                                                                            
go test runtime/cgo: open /snap/go/2130/pkg/linux_amd64/runtime/cgo.a: read-only file system
```

Here's the `ginkgo build -n` output:
```
$ ginkgo build -n | head -n 5
Compiling test...
mkdir -p $WORK/b041/
mkdir -p /snap/go/2130/pkg/linux_amd64/runtime/
mv /home/joe/.cache/go-build/fa/fab0bd7339dd2fd637536e97a5820b132da831d8fae3059a7717544d69870a1e-d /snap/go/2130/pkg/linux_amd64/runtime/cgo.a
mkdir -p $WORK/b036/
```

Basically it seems like the ginkgo build is attempting to overwrite my Golang installation.

I believe that this is because `-i` is added to the `go test` arguments here:

https://github.com/onsi/ginkgo/blob/39febac9157b63aeba74d843d6b1c30990f3d7ed/ginkgo/testrunner/test_runner.go#L67

In Golang 1.10, `-i` should no longer be required:

https://golang.org/doc/go1.10#build

> The old advice to add the -i flag for speed, as in go build -i or go test -i, is no longer necessary: builds run just as fast without -i. For more details, see go help cache.

Here's the related upstream issue:

https://github.com/golang/go/issues/24674

When I applied this diff and installed, I could run ginkgo again:

```
$ git di
diff --git a/ginkgo/testrunner/test_runner.go b/ginkgo/testrunner/test_runner.go
index 5df41be4c72e..f45de78ce6de 100644
--- a/ginkgo/testrunner/test_runner.go
+++ b/ginkgo/testrunner/test_runner.go
@@ -64,7 +64,7 @@ func (t *TestRunner) Compile() error {
 }
 
 func (t *TestRunner) BuildArgs(path string) []string {
[...]

PR #505, opened 2018-07-16
Fix ginkgo run with Golang 1.10 snap archive on Ubuntu 18.04

joestringer:
Since Golang 1.10, this option should no longer be required:

https://golang.org/doc/go1.10#build

> The old advice to add the -i flag for speed, as in go build -i or go
test -i, is no longer necessary: builds run just as fast without -i. For
more details, see go help cache.

Furthermore, it creates issues like the below when running ginkgo
against a snap-installed version of Golang 1.10 on Ubuntu 18.04:

```
$ ginkgo build
Compiling test...
Failed to compile test:

go test runtime/cgo: open /snap/go/2130/pkg/linux_amd64/runtime/cgo.a: read-only file system
```

Fixes: #504

### DIFF
--- a/ginkgo/testrunner/build_args.go
+++ b/ginkgo/testrunner/build_args.go
@@ -1,0 +1,7 @@
+// +build go1.10
+
+package testrunner
+
+var (
+	buildArgs = []string{"test", "-c"}
+)
--- a/ginkgo/testrunner/build_args_old.go
+++ b/ginkgo/testrunner/build_args_old.go
@@ -1,0 +1,7 @@
+// +build !go1.10
+
+package testrunner
+
+var (
+	buildArgs = []string{"test", "-c", "-i"}
+)
--- a/ginkgo/testrunner/test_runner.go
+++ b/ginkgo/testrunner/test_runner.go
@@ -64,7 +64,9 @@ func (t *TestRunner) Compile() error {
 }
 
 func (t *TestRunner) BuildArgs(path string) []string {
-	args := []string{"test", "-c", "-i", "-o", path, t.Suite.Path}
+	args := make([]string, len(buildArgs), len(buildArgs)+3)
+	copy(args, buildArgs)
+	args = append(args, "-o", path, t.Suite.Path)
 
 	if t.getCoverMode() != "" {
 		args = append(args, "-cover", fmt.Sprintf("-covermode=%s", t.getCoverMode()))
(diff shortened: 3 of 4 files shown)


nodo:
Thanks @joestringer ! 

I am a bit concerned in removing the `-i` flag, since we are currently supporting Go 1.6. Also, I have tried to reproduce the steps and seem very different from yours. Could this be an issue with Snap?

joestringer:
Hi @nodo!

I can definitely appreciate that, Golang 1.10 is fairly new. 

For a bit more information, the maintainer of the Snap package described part of the issue here (some quotes below):

https://github.com/golang/go/issues/24674#issuecomment-379919460

> For the snap packaging, I do know why runtime/cgo appears stale. It's fairly obscure though: the snap packaging builds runtime/cgo with particular C compiler flags to ensure that the generated .o files can be handled by older compilers. The go tool now keeps track of these, so unless you happen to use exactly the same flags, the go tool considers runtime/cgo to be stale.

Some others on the same thread suggested that people should just move away from `-i`, which makes sense if you can guarantee Golang 1.10.

Perhaps a better alternative would be to be able to parameterize this option rather than hardcoding it? By default, it could always add `-i`, then for users on newer Golang already they set some config option somewhere to disable the `-i`, or alias something like `ginkgo=ginkgo --build-opts-i=false`?

joestringer:
@ueokande Perhaps then there's a better way that uses something like https://golang.org/pkg/runtime/#Version to detect and inject this at runtime.

/cc @nodo thoughts?

ymmt2005:
It would be better to use build constraints:
https://golang.org/pkg/go/build/#hdr-Build_Constraints

To have different build args for Go <1.10 and >=1.10, create these two files:

`build_args.go`:
```go
// +build go1.10

var buildArgs = []string{"test", "-c"}
```

`build_args_old.go`:
```go
// +build !go1.10

var buildArgs = []string{"test", "-c", "-i"}
```

and reference `buildArgs` like this:
```go
func (t *TestRunner) BuildArgs(path string) []string {
    args := make([]string, len(buildArgs), len(buildArgs)+3)
    copy(args, buildArgs)
    args = append(args, "-o", path, t.Suite.Path)
    ...
}
```

joestringer:
Thanks @ymmt2005, I used your approach.

joestringer:
@nodo did you get a chance to take a look?

nodo:
Hey @joestringer - looks good to me I have a few doubts about the testing, but I cannot think about a better solution for now 😄 

Thanks a lot for contributing! 💯

joestringer:
@nodo Thanks! I agree it's not the most elegant, but I couldn't think of something better.

Fortunately I think if the testing changes significantly and this breaks, it should err on the side of "breaking when it shouldn't" rather than "hiding a breakage when it *should* break".

ymmt2005:
Thanks!

Just for your information, you can use `buildArgs` to construct the expected value in testing.

PR #506, opened 2018-07-17
Minor doc and go vet fixes

carldunham:
Split out from #502.


nodo:
🎉  thanks @carldunham !

PR #527, opened 2018-10-09
Add Go 1.11 to Travis build matrix

emirb:


blgm:
Various problems with Ginkgo and Go v1.11 have been reported, but adding this to Travis will make sure that we will notice if things get worse.  Thanks for your contribution @emirb!

Issue #529, opened 2018-10-26
ginkgo CLI cannot run pre-compiled test package on windows


olegch:
# Reproduce

Compile ginkgo for windows:
```
GOOS=windows GOARCH=amd64 go build -i -v -o ginkgo.exe github.com/onsi/ginkgo/ginkgo
```
Precompile test package for windows:
```
GOOS=windows GOARCH=amd64 go test -i -v -c -o target/my.test.exe my-test-package
```
Try running the test binary on windows:
```
my.test.exe
```
It works as expected.
Try running the test binary via ginkgo CLI:
```
ginkgo.exe my.test.exe
```
Test suite is not found and not run

olegch:
It looks like the heart of the problem lies here: https://github.com/onsi/ginkgo/blob/master/ginkgo/testsuite/test_suite.go#L29-L35

Ginkgo assumes that the precompiled test binary name must end with `.test` and that it must have an executable bit set.

Neither is true on windows.

I tried to remove these checks from the code and apparently it fixes the issue, at least partially.
With those checks removed `ginkgo`  can run precompiled test package on windows correctly, but only if the binary file does not have `.exe` extension.
It is not ideal (you want to have `exe` extension on windows to run the file), but it works.

I suppose it is possible to fix ginkgo to accept binary names ending with `exe`, but it requires fixes in a couple more places.

onsi:
I'm working through the backlog of old Ginkgo issues - apologies as this issue is probably stale now.

I think you've honed in on why this is happening - I'll fix it and ensure it works in the v2 work I'm doing.

I'm going to close this for now, but feel free to reopen it if you'd like.

PR #531, opened 2018-10-27
fix running precompiled tests via ginkgo CLI on windows (#529)

olegch:
Hello,

Here is a suggested change to support running precompiled tests on windows via ginkgo CLI (see issue #529 for more details).

### DIFF
--- a/ginkgo/testsuite/test_suite.go
+++ b/ginkgo/testsuite/test_suite.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


ginkgo/testsuite/test_suite.go
```diff
@@ -26,16 +27,22 @@ func PrecompiledTestSuite(path string) (TestSuite, error) {
...
+	if runtime.GOOS != "windows" && info.Mode()&0111 == 0 {
 		return TestSuite{}, errors.New("this is not executable")
 	}
 
 	dir := relPath(filepath.Dir(path))
-	packageName := strings.TrimSuffix(filepath.Base(path), filepath.Ext(path))
+	var packageName string
+	if strings.HasSuffix(path, ".test.exe") {
+		packageName = strings.TrimSuffix(filepath.Base(path), ".test.exe")
+	} else {
```
wingyplus: Can we move assign value in else block to `packageName` declaration?

blgm:
One thing that could be improved is to use Go standard library functions that automatically append `.exe` on Windows. For instance when [the `exec.LookPath` function](https://golang.org/pkg/os/exec/#LookPath) is called with a path name containing a `/` (or `\` on Windows) it will look for files with a `.exe` extension. The advantage of this is separation of concerns, and also less code to maintain.

johnSchnake:
Checking on the status of this. Happy to help if needed.

I had originally thought that just getting rid of the exe suffix would have been sufficient but was surprised in my testing to find thats not the case.

PR #560, opened 2019-01-25
Optimize focus to avoid allocations

smarterclayton:
Both regex creation and ConcatenatedString() are expensive. Memoize
them within the spec iterator so we don't recalculate them later.

Found this when working with large sets of tests (>10k) within a
single binary.

### DIFF
--- a/internal/spec/specs.go
+++ b/internal/spec/specs.go
@@ -7,14 +7,21 @@ import (
 )
 
 type Specs struct {
-	specs                []*Spec
+	specs []*Spec
+	names []string
+
 	hasProgrammaticFocus bool
 	RegexScansFilePath   bool
 }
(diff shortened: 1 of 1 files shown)


williammartin:
Thanks for the PR, unfortunately it looks like it has broken a test :(

smarterclayton:
Fixed, sorry (typo'd skipFilter and focusFilter)

williammartin:
Quick question for curiosity, for your large set of tests, what is the performance difference for having this PR and not?

smarterclayton:
Something on the order of a million object allocations or so to get to the point where we run no tests, vs 50k or so.

williammartin:
Cool cool, well LGTM!

Issue #559, opened 2019-01-25
Support for vendoring + modules


zmackie:
The team I'm on vendors (for reasons...) but we're also using modules. It'd be great if ginkgo's pre-test compile step could respect the existence of this case and run the build with local code (ala `go build -mod=vendor`. We'd be happy about contributing a PR given a bit of guidance on where to look...

BooleanCat:
As of ginkgo 1.8.0, you can use `ginkgo -mod vendor ...`

williammartin:
Thanks @BooleanCat for the contribution.

@Zanadar Can you check and see if this solves your issue?

Zanadar:
Thanks @booleancat

Zanadar:
@williammartin sure I’ll check. Will re-close if we’re all set.

Zanadar:
@williammartin @BooleanCat 
I'm not certain this is actually working...it looks like ginkgo is still downloading modules that exist in our local vendor folder

Eg

```
± dt |develop {17} U:16 ?:68 ✗| → docker run -v $(pwd):/app -it golang:1.12.1 bash
root@d95aa2cffaba:/go# cd /app/
root@d95aa2cffaba:/app# ./scripts/build.sh
++ dirname ./scripts/build.sh
+ cd ./scripts/..
+ source .envrc
+++ pwd
++ export GOBIN=/app/.bin
++ GOBIN=/app/.bin
++ export PATH=/app/.bin:/go/bin:/usr/local/go/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
++ PATH=/app/.bin:/go/bin:/usr/local/go/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+ GOOS=linux
+ go build -mod=vendor '-ldflags=-s -w' -o bin/supply ./src/nodejs/supply/cli
+ GOOS=linux
+ go build -mod=vendor '-ldflags=-s -w' -o bin/finalize ./src/nodejs/finalize/cli
root@d95aa2cffaba:/app#
```
(notice modules do not need to be downloaded)
vs

```
root@d95aa2cffaba:/app# ./scripts/unit.sh
++ dirname ./scripts/unit.sh
+ cd ./scripts/..
+ source .envrc
+++ pwd
++ export GOBIN=/app/.bin
++ GOBIN=/app/.bin
[...]

PR #565, opened 2019-03-07
Support testing projects using go modules with vendor directories

BooleanCat:
Currently when testing a project using a vendor directory and go modules, I see that ginkgo is unaware of the `-mod` compilation flag given to the go binary. Since it is never given, dependencies are always pulled instead of being found in vendor/.

This change will pass a `-mod` flag through to the go build command.


BooleanCat:
Related to #559

PR #571, opened 2019-04-10
Print Skip reason in JUnit reporter if one was provided

dollierp:


williammartin:
This seems reasonable but it would be good to get some test coverage.

PR #575, opened 2019-04-20
Update index.md, reduce the redundant block

linzhaoming:


williammartin:
LGTM thanks for the spot!

PR #577, opened 2019-05-20
Print help to stdout when the user asks for help

### LINKED ISSUE #567: Help should be printed to stdout
`ginkgo -h` produces a lot of output. Sometimes I am looking for something and grepping doesn't help me unless I redirect stderr to stdout. It would be really nice if the help was printed to stdout.

akshaymankar:
This changes behaviour of `ginkgo -h`. Before this PR, it used to print a different usage from what `ginkgo help` prints. This is caused by go's inbuilt flag library choosing the print the `defaultUsage` when the `Usage` func is not defined on a `flagSet`.

Another thing to note is `ginkgo -h` exits with exit code 2, `ginkgo help` exits with 0. This is consistent with previous behaviour. Ideally we should change `ginkgo -h` to exit with 0, but it is currently hard-coded in the flag library.

PS: `ginkgo --help` behaves exactly the same as `ginkgo -h`.

[Fixes #567]

### DIFF
--- a/ginkgo/main.go
+++ b/ginkgo/main.go
@@ -153,6 +153,7 @@ func (c *Command) Matches(name string) bool {
 }
 
 func (c *Command) Run(args []string, additionalArgs []string) {
+	c.FlagSet.Usage = usage
 	c.FlagSet.Parse(args)
 	c.Command(c.FlagSet.Args(), additionalArgs)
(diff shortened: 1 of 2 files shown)


williammartin:
LGTM thanks

PR #578, opened 2019-05-22
Add support for go module when running `generate` command

### LINKED ISSUE #524: ginkgo generate fails in module
It seems that running `ginkgo generate` inside a go module, as per go 1.11, somewhere outside `GOPATH` results in a broken test skeleton.

Example:

```go
package something_test

import (
	. "github.com/onsi/ginkgo"
	. "github.com/onsi/gomega"

	. "UNKNOWN_PACKAGE_PATH"
)

var _ = Describe("Something", func() {

})
```

The problem likely stems from the usage of `ImportDir` which does not work with go modules:
https://github.com/onsi/ginkgo/blob/000d317071340972710dc1da9ae3c2c016a7a560/ginkgo/bootstrap_command.go#L106

This has been described here:
https://github.com/golang/go/issues/26504

A workaround is to use the [golang.org/x/tools/go/packages](https://godoc.org/golang.org/x/tools/go/packages) package, however it is not stable yet.

leninhasda:
Add support for go module when running `generate` command
Typo fix

This changes how `generate` command looks for import path. Now it will
try to find `go.mod` file and determine import path from there first,
and will fallback to old way otherwise.

Ref:
Two functions I used are from original go source code:
* [findModuleRoot](https://github.com/golang/go/blob/master/src/cmd/go/internal/modload/init.go#L480)
* [moduleName ~ slightly modified of ModulePath](https://github.com/golang/go/blob/master/src/cmd/go/internal/modfile/read.go#L837)

[Fixes #524]

### DIFF
--- a/ginkgo/generate_command.go
+++ b/ginkgo/generate_command.go
@@ -1,10 +1,12 @@
 package main
 
 import (
+	"bytes"
 	"flag"
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"text/template"
 )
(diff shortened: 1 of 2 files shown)


kvptkr:
@onsi Who would be best to review this? This is a pain point now that modules are GA

blgm:
Hi @kvptkr @leninhasda. The implementation here looks good and has worked for me when I have tried it out. The thing that this PR is missing is some tests, which could live here: https://github.com/onsi/ginkgo/blob/master/integration/subcommand_test.go#L16

blgm:
@leninhasda, thank you for this contribution. It's a valuable improvement. I have added a test for it in commit 8892a1d7a3b76ea9fa1eb4c08788dd8175460e21 as this will hopefully reduce the chance of this being accidentally broken in the future. If you can see any potential improvement to the testing, then please feel free to submit another PR.

Issue #582, opened 2019-06-19
GinkgoT() should implements testing.TB


wingyplus:
I found that I cannot parse `GinkgoT()` as an argument to my test helper code that use both testing and benchmark code because it doesn't have `Helper()` method. 

For example: 

```
package gink

import (
	"testing"

	"github.com/onsi/ginkgo"
	"github.com/onsi/gomega"
)

func Test(t *testing.T) {
	gomega.RegisterFailHandler(ginkgo.Fail)
	ginkgo.RunSpecs(t, "suite")
}

var _ = ginkgo.Describe("my suite", func() {
	ginkgo.It("under test", func() {
		myTestHelperCode(ginkgo.GinkgoT())
	})
})

// I use this function across my test code and benchmark.
func myTestHelperCode(tb testing.TB) {
	tb.Helper()
}
```

The result from `go test`:

```
$ go test .
# github.com/wingyplus/gink [github.com/wingyplus/gink.test]
./gink_test.go:17:34: cannot use ginkgo.GinkgoT() (type ginkgo.GinkgoTInterface) as type testing.TB in argument to myTestHelperCode:
	ginkgo.GinkgoTInterface does not implement testing.TB (missing Helper method)
FAIL	github.com/wingyplus/gink [build failed]
FAIL
```

jmalloc:
Hi all, 

Apologies in advance for the wall-of-text!

I too have been looking for functionality along these lines. I am hopeful that you might (re)consider updating `GinkgoTInterface` to be compatible with `testing.TB`. 

I figured it was also worth pointing out that `testing.TB` itself contains an unexported method, and as such can never be implemented by anything outside the `testing` package 😞, though I don't believe that invalidates the argument for supporting such a change.

My use case is that I maintain a test runner / assertion library for use with a particular application framework. It has it's own equivalent to `GinkgoTInterface` which allows it to be used under Ginkgo, the standard test runner, or anything else that can supply such an interface.

This assertion library also supports custom user-supplied assertions. I would like to be able to supply something compatible with `testing.TB` to my users such that their custom assertions could make use of any other package that uses its own subset of `TB`.

We use Ginkgo and Gomega widely within my organisation, and as such I have been reluctant to require any methods that Ginkgo can not supply, but this is becoming increasingly more difficult as I seek to allow integration with third-party testing/assertion libraries that are unknown to me.

Obviously this is just one scenario, but I hope it makes the case for a generalised approach to supporting this.

[...]

Issue #563, opened 2019-02-24
Add "Helper()" to GinkgoTInterface


shaunc:
`https://github.com/bradleyjkemp/cupaloy` uses [Helper()](https://godoc.org/testing#T.Helper) could this be added to the mock?

shaunc:
Also is missing "Name()". As a rough workaround, the following seems to work:
```
type TestingT struct {
	GinkgoTInterface
	desc GinkgoTestDescription
}

func NewTestingT() TestingT {
	return TestingT{GinkgoT(), CurrentGinkgoTestDescription()}
}

func (i TestingT) Helper() {

}
func (i TestingT) Name() string {
	return i.desc.FullTestText
}
```

blgm:
It looks like this has already been implemented in commit https://github.com/onsi/ginkgo/commit/b5fe44d9e12817942846c49d35804c5100e05327 so I'm going to close this. Please let me know if I missed something.

PR #585, opened 2019-06-25
Add Helper() method to GinkgoTInterface

bstick12:
* This aligns more with the testing.TB interface

This has been requested in 
* https://github.com/onsi/ginkgo/issues/582
* https://github.com/onsi/ginkgo/issues/563. 

I was also looking for the functionality to use https://github.com/cloudfoundry/libcfbuildpack

### DIFF
--- a/ginkgo_dsl.go
+++ b/ginkgo_dsl.go
@@ -120,6 +120,7 @@ type GinkgoTInterface interface {
 	Skipf(format string, args ...interface{})
 	SkipNow()
 	Skipped() bool
+	Helper()
 }
 
 //Custom Ginkgo test reporters must implement the Reporter interface.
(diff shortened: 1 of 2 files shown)


williammartin:
I'm not totally sure about this. There is an expectation from tests is that `Helper()` results in some behaviour around error reporting similar to `ExpectWithOffset`. That said, we have an equivalent that does nothing in `Parallel()`.

I'd be inclined to merge to help people out but I could imagine it causing confusion in the future.

What do you think @blgm @ansd ?

jmalloc:
Even if `Helper()` was a no-op, I too would find this very useful from an integration standpoint.

jjaferson:
Hey guys,

Any idea whether this PR is going to get merged? 

it would be nice to have `GinkgoTInterface` compatible with `testing.TB` even if `Helper()` is a no-op

blgm:
Hi @jjaferson. It looks like this has already been implemented in commit https://github.com/onsi/ginkgo/commit/b5fe44d9e12817942846c49d35804c5100e05327

Ideally this PR would have been closed before, but since the code has already been merged in another PR, I don't think there's much value in keeping it around. Please let me know if there's something I missed.

jjaferson:
@blgm thanks, I was using an old version of GinkGo I noticed it's been added in 1.14.2.

Issue #583, opened 2019-06-25
Option to print output of passed tests in Reporters


manosnoam:
For those who use reporters, such as Teamcity or Junit, for their tests frameworks, it is commonly required to display steps output of passed tests (not only for failed tests, which is the current behavior).

I've implemented an option to use `-ginkgo.reportPassed ` for that purpose, 
and the report output, for example in Junit XML, will look as follow:

```
<?xml version="1.0" encoding="UTF-8"?>
  <testsuite name="E2E suite" tests="1" failures="0" errors="0" time="11.396">
      <testcase name="list existing nodes on the cluster" classname="E2E suite" time="0">
          <skipped></skipped>
      </testcase>
      <testcase name="Pod to Service tests" classname="E2E suite" time="11.395812505">
          <passed>STEP: Creating kubernetes clients
          STEP: Building namespace api objects, basename dataplane-p2s-nd1
          STEP: Creating a namespace e2e-tests-dataplane-p2s-nd1-q5phj to execute the test 
          STEP: Creating a listener pod in cluster B, which will wait for a handshake over TCP
          STEP: Pointing a service ClusterIP to the listerner pod in cluster; Jun 25 15:23:54.494: INFO: Service for listener pod has ClusterIP: 100.96.40.208
          STEP: Creating a connector pod in cluster B, which will attempt the specific UUID handshake over TCP;</passed>
      </testcase>
      <testcase name="create a pod using the provided client" classname="E2E suite" time="0">
      <failure type="Failure">/home/nmanos/go/src/github.com/rancher/submariner/test/e2e/dataplane/tcp_pod_to_service.go:14;Expected;    ;string: 100.96.190.228 (100.96.190.228:1234) open;    ;to contain substring;    ;string: 7277d684-2c95-40fd-9bb5-5bf791a15ac7;/home/nmanos/go/src/github.com/rancher/submariner/test/e2e/dataplane/tcp_pod_to_service.go:46</failure>
      <system-out>STEP: Creating kubernetes clients
          STEP: Building namespace api objects, basename dataplane-p2s-nd
          STEP: Creating a namespace e2e-tests-dataplane-p2s-nd-ttds8 to execute the test</system-out>
      </testcase>
  </testsuite>
```

manosnoam:
Hi @williammartin,
I've implemented your suggested requirements in https://github.com/onsi/ginkgo/pull/586, 
including Tests for:

- internal/specrunner/spec_runner_test.go
[...]

PR #586, opened 2019-06-25
Option to print output of passed tests in Reporters

manosnoam:
This is PR for issue #583

A new CLI option: -ginkgo.reportPassed  :
It will print output for each passed test in the generated report,
including JUnit, Teamcity, and Default reporters.

For example, in JUnit (XML), the test output will be added under:
`<testcase> <passed>`

The default behavior (without this option), prints test output
only if the test case (spec) has failed.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -52,6 +52,7 @@ type DefaultReporterConfigType struct {
 	Succinct          bool
 	Verbose           bool
 	FullTrace         bool
+	ReportPassed      bool
 }
 
 var DefaultReporterConfig = DefaultReporterConfigType{}
@@ -98,6 +99,7 @@ func Flags(flagSet *flag.FlagSet, prefix string, includeParallelFlags bool) {
 	flagSet.BoolVar(&(DefaultReporterConfig.Verbose), prefix+"v", false, "If set, default reporter print out all specs as they begin.")
 	flagSet.BoolVar(&(DefaultReporterConfig.Succinct), prefix+"succinct", false, "If set, default reporter prints out a very succinct report")
 	flagSet.BoolVar(&(DefaultReporterConfig.FullTrace), prefix+"trace", false, "If set, default reporter prints out the full stack trace when a failure occurs")
+	flagSet.BoolVar(&(DefaultReporterConfig.ReportPassed), prefix+"reportPassed", false, "If set, default reporter prints out captured output of passed tests.")
 }
 
 func BuildFlagArgs(prefix string, ginkgo GinkgoConfigType, reporter DefaultReporterConfigType) []string {
@@ -194,6 +196,10 @@ func BuildFlagArgs(prefix string, ginkgo GinkgoConfigType, reporter DefaultRepor
 
 	if reporter.FullTrace {
 		result = append(result, fmt.Sprintf("--%strace", prefix))
+	}
+
+	if reporter.ReportPassed {
+		result = append(result, fmt.Sprintf("--%sreportPassed", prefix))
 	}
 
 	return result
--- a/internal/specrunner/spec_runner.go
+++ b/internal/specrunner/spec_runner.go
@@ -300,7 +300,7 @@ func (runner *SpecRunner) reportSpecWillRun(summary *types.SpecSummary) {
 }
 
 func (runner *SpecRunner) reportSpecDidComplete(summary *types.SpecSummary, failed bool) {
-	if failed && len(summary.CapturedOutput) == 0 {
+	if len(summary.CapturedOutput) == 0 {
 		summary.CapturedOutput = string(runner.writer.Bytes())
 	}
 	for i := len(runner.reporters) - 1; i >= 1; i-- {
--- a/internal/specrunner/spec_runner_test.go
+++ b/internal/specrunner/spec_runner_test.go
@@ -720,6 +720,7 @@ var _ = Describe("Spec Runner", func() {
 				"R1.WillRun",
 				"R2.WillRun",
 				"A",
+				"BYTES",
 				"R2.DidComplete",
 				"R1.DidComplete",
 				"TRUNCATE",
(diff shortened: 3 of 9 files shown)


manosnoam:
Hi @williammartin,
I saw you've reviewed #571 (Print Skip reason in JUnit reporter if one was provided), 
so it'll be great if you could review this PR as well. 
I've added this feature test into reporters/default_reporter_test.go

Please see issue #583 for more details.
Thanks!

williammartin:
Hi @manosnoam,

I think the general idea of enabling `passed` to be included in the output makes sense, thanks.

I think there's probably some missing tests in `junit_reporter_test.go` that covers the additional logic in `junit_reporter.go`. As far as I can see here the only test is that the flag gets passed through.

That said, I don't love the idea of introducing a reporter specific flag. I wonder whether any of the following options are better:
 - Make use of the existing `-v` verbose flag (has slightly different semantics but similar outcome)
 - Provide a new flag `-passed` which would work for all reporters (would have to implement for all reporters)
 - Make this a compile time configuration (would not be toggleable and wouldn't work with old test suites) e.g.

```	
junitReporter := reporters.NewJUnitReporterWithConfig("junit.xml", { passed: true })
```

I think there are advantages and disadvantages to all of these options. Let me know your thoughts.

manosnoam:
@williammartin, thanks for super fast reply!
I think `-v` (verbose) is usually for detailed console output, but not for the generated reports.
`-passed` is what I've implemented, but I called it `-junitPassed`.

I guess there's a way to add test of this flag, can you suggest how to test this CLI flag ?

williammartin:
> I think -v (verbose) is usually for detailed console output, but not for the generated reports.

I think these are maybe just two sides of the same coin. The console output is the default reporter that you can combine with other reporters. What I'm suggesting is maybe `-v` should be used to signify the same thing (but I'm not convinced they are exactly equivalent which is why I might favour `-passed`).

> -passed is what I've implemented, but I called it -junitPassed

yeh I think this what bothers me though, for example when I look at https://github.com/onsi/ginkgo/pull/586/files#diff-245d401c0f93bef3b7927b60d3418905L303 there's no reason that a reporter abstraction should leak into this thing that knows how to run specs.

Similarly, this is behaviour that should probably not be _solely_ useful for the junit reporter but also for default or teamcity (or any other custom reporter). 

> I guess there's a way to add test of this flag, can you suggest how to test this CLI flag ?

Ideally we would have an integration test that actually exercises a test suite using this feature, but at the very least I would expect to see something in https://github.com/onsi/ginkgo/blob/master/reporters/junit_reporter_test.go#L79-L92 that ensures a `PassedMessage` is set.

Hope this helps.

manosnoam:
- What about renaming CLI flag to `-reportPassed`, and in `config/config.go` to `ReportPassed bool` ?
- In `specrunner/spec_runner.go` you right, we don't really need to interact with `DefaultReporterConfig`.
The functionality works without it, so all we need is to append output to the summary:
`if len(summary.CapturedOutput) == 0 { 
		summary.CapturedOutput = string(runner.writer.Bytes())
	}`
- Actually, the flag is already evaluated in `juinit_reporter.go`: https://github.com/onsi/ginkgo/pull/586/files#diff-87479b8a1ad5040d4d2558e52c89cf07R115
I'll add same condition check to `teamcity_reporter.go` and `default_reporter.go`
- I will add integration tests to these as well.

manosnoam:
Hi @williammartin,
I've implemented the requirments from my last comment.
However, [Travis-CI](https://travis-ci.org/onsi/ginkgo/builds/551542525) seems to be failing on: 
```

[1561675640] Spec Runner Suite - 55/55 specs ••••••••••••••••••••••••
------------------------------
• Failure [0.019 seconds]
Spec Runner Managing the writer [It] should truncate between tests, but only dump if a test fails 
/home/travis/gopath/src/github.com/onsi/ginkgo/internal/specrunner/spec_runner_test.go:717
  Expected
      <[]string | len:22, cap:32>: ["TRUNCATE", "R1.WillRun", "R2.WillRun", "A", "BYTES", "R2.DidComplete", "R1.DidComplete", "TRUNCATE", "R1.WillRun", "R2.WillRun", "B", "BYTES", "R2.DidComplete", "DUMP", "R1.DidComplete", "TRUNCATE", "R1.WillRun", "R2.WillRun", "C", "BYTES", "R2.DidComplete", "R1.DidComplete"]
  to equal
      <[]string | len:20, cap:20>: ["TRUNCATE", "R1.WillRun", "R2.WillRun", "A", "R2.DidComplete", "R1.DidComplete", "TRUNCATE", "R1.WillRun", "R2.WillRun", "B", "BYTES", "R2.DidComplete", "DUMP", "R1.DidComplete", "TRUNCATE", "R1.WillRun", "R2.WillRun", "C", "R2.DidComplete", "R1.DidComplete"]
  /home/travis/gopath/src/github.com/onsi/ginkgo/internal/specrunner/spec_runner_test.go:718
  Full Stack Trace
  	/home/travis/gopath/src/github.com/onsi/gomega/internal/assertion/assertion.go:75 +0x2e2
  github.com/onsi/gomega/internal/assertion.(*Assertion).Should(0xc00046b340, 0xb47c60, 0xc0001c32e0, 0x0, 0x0, 0x0, 0x0)
  	/home/travis/gopath/src/github.com/onsi/gomega/internal/assertion/assertion.go:28 +0x11c
  github.com/onsi/ginkgo/internal/specrunner_test.glob..func1.14.2()
  	/home/travis/.gimme/versions/go1.12.6.linux.amd64/src/testing/testing.go:865 +0x164
  created by testing.(*T).Run
  	/home/travis/.gimme/versions/go1.12.6.linux.amd64/src/testing/testing.go:916 +0x65b

Any chance you see what the culprit ?

```

williammartin:
I had a play around and it seems to make sense to me. Thanks!

williammartin:
I think there's something funny going on with the commits in this PR though, can you rebase to just one commit describing this change?

manosnoam:
> I think there's something funny going on with the commits in this PR though, can you rebase to just one commit describing this change?

Yep strange, got this fixed and rebased.  Thanks!
[CI passed](https://travis-ci.org/onsi/ginkgo/builds/555217686).

PR #590, opened 2019-07-17
stack backtrace: fix alignment and skipping

pohly:
With go 1.12.6, runtime.Stack() returns a stack that includes one
additional line with the goroutine and that has runtime.Stack itself
as first entry.

PruneStack then returned a pruned stack that started with a source
code line instead of the function name and that did not skip all
intermediate entries.

Example:

 Stack:
	"runtime/debug.Stack(0x0, 0x0, 0x0)",
	"\t/nvme/gopath/go/src/runtime/debug/stack.go:24 +0xa1",
	"k8s.io/kubernetes/vendor/github.com/onsi/ginkgo/internal/codeloc...+39 more",
	"\t/nvme/gopath/src/k8s.io/kubernetes/vendor/github.com/onsi/ginkg...+49 more",
	"k8s.io/kubernetes/vendor/github.com/onsi/ginkgo.Fail(0xc00052677...+32 more",
	"\t/nvme/gopath/src/k8s.io/kubernetes/vendor/github.com/onsi/ginkg...+25 more",
	"k8s.io/kubernetes/test/e2e/framework/ginkgowrapper.Fail(0xc00052...+35 more",
	"\t/nvme/gopath/src/k8s.io/kubernetes/test/e2e/framework/ginkgowra...+25 more",
	"k8s.io/kubernetes/test/e2e/framework/log.Fail(0xc00038e1e0, 0x4f...+25 more",
	"\t/nvme/gopath/src/k8s.io/kubernetes/test/e2e/framework/log/logge...+14 more",
	"k8s.io/kubernetes/vendor/github.com/onsi/gomega/internal/asserti...+103 more",
	"\t/nvme/gopath/src/k8s.io/kubernetes/vendor/github.com/onsi/gomeg...+43 more",
	"k8s.io/kubernetes/vendor/github.com/onsi/gomega/internal/asserti...+95 more",
	"\t/nvme/gopath/src/k8s.io/kubernetes/vendor/github.com/onsi/gomeg...+42 more",
	"k8s.io/kubernetes/test/e2e/framework/log_test.glob..func1.5()",
        ...
 Skip: 5

 Resulting stack (for some other example):
            /nvme/gopath/src/k8s.io/kubernetes/test/e2e/framework/log/logger.go:51 +0x143
      k8s.io/kubernetes/test/e2e/framework/log.Failf(...)
            /nvme/gopath/src/k8s.io/kubernetes/test/e2e/framework/log/logger.go:43
=>    k8s.io/kubernetes/test/e2e/framework/log_test.glob..func1.2.1(...)
            /nvme/gopath/src/k8s.io/kubernetes/test/e2e/framework/log/logger_test.go:41
      k8s.io/kubernetes/test/e2e/framework/log_test.glob..func1.2()
            /nvme/gopath/src/k8s.io/kubernetes/test/e2e/framework/log/logger_test.go:42 +0x52
      k8s.io/kubernetes/vendor/github.com/onsi/ginkgo/internal/leafnodes.(*runner).runSync(0xc00029b020, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
            /nvme/gopath/src/k8s.io/kubernetes/test/e2e/framework/log/logger_test.go:65 +0x1c8
      testing.tRunner(0xc000358600, 0x19818c0)
            /nvme/gopath/go/src/testing/testing.go:865 +0xc0
      created by testing.(*T).Run
            /nvme/gopath/go/src/testing/testing.go:916 +0x35a

It should start with the line marked with =>. To achieve that, both
the goroutine line and the runtime.Stack lines must be skipped. The
code does that with some sanity checks to ensure that this also works
with Go version where those lines (presumably) weren't part of the
stack.

### DIFF
--- a/internal/codelocation/code_location.go
+++ b/internal/codelocation/code_location.go
@@ -15,15 +15,31 @@ func New(skip int) types.CodeLocation {
 	return types.CodeLocation{FileName: file, LineNumber: line, FullStackTrace: stackTrace}
 }
 
+// PruneStack removes references to functions that are internal to Ginkgo
+// and the Go runtime from a stack string and a certain number of stack entries
+// at the beginning of the stack. The stack string has the format
+// as returned by runtime/debug.Stack. The leading goroutine information is
+// optional and always removed if present. Beware that runtime/debug.Stack
+// adds itself as first entry, so typically skip must be >= 1 to remove that
+// entry.
 func PruneStack(fullStackTrace string, skip int) string {
 	stack := strings.Split(fullStackTrace, "\n")
+	// Ensure that the even entries are the method names and the
+	// the odd entries the source code information.
+	if len(stack) > 0 && strings.HasPrefix(stack[0], "goroutine ") {
+		// Ignore "goroutine 29 [running]:" line.
+		stack = stack[1:]
+	}
+	// The "+1" is for skipping over the initial entry, which is
+	// runtime/debug.Stack() itself.
 	if len(stack) > 2*(skip+1) {
 		stack = stack[2*(skip+1):]
 	}
 	prunedStack := []string{}
 	re := regexp.MustCompile(`\/ginkgo\/|\/pkg\/testing\/|\/pkg\/runtime\/`)
 	for i := 0; i < len(stack)/2; i++ {
-		if !re.Match([]byte(stack[i*2])) {
+		// We filter out based on the source code file name.
+		if !re.Match([]byte(stack[i*2+1])) {
 			prunedStack = append(prunedStack, stack[i*2])
 			prunedStack = append(prunedStack, stack[i*2+1])
 		}
--- a/internal/codelocation/code_location_test.go
+++ b/internal/codelocation/code_location_test.go
@@ -1,7 +1,10 @@
 package codelocation_test
 
 import (
+	"fmt"
 	"runtime"
+	"runtime/debug"
+	"strings"
 
 	. "github.com/onsi/ginkgo"
 	"github.com/onsi/ginkgo/internal/codelocation"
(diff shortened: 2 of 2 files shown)


neolit123:
@onsi
hi, could you please take a look at this PR, we are seeing the problem in k8s upstream.
thanks!

blgm:
Hi @neolit123, thank you for submitting a PR
- I can see that TravisCI is failing for this PR. What would it take to get the tests to pass?
- What are the implications for earlier versions of Go?
- The PruneStack function has tests. Is it possible to extend those tests, or add some testing elsewhere?

neolit123:
^ @pohly could you please look at the failing CI and answering the questions @blgm poses.

pohly:
> I can see that TravisCI is failing for this PR. What would it take to get the tests to pass?

I broke pruning when shifting the lines by one. The actual fix needs to be a bit more complex. It's working now, please have another look.

I also found that it is Kubernetes, not Ginkgo, which might be off by one in its skip count. We need to discuss in Kubernetes whether `framework.ExpectError` should include itself in the full stack trace - IMHO not. But that doesn't belong here.

> What are the implications for earlier versions of Go?

I thought that this was some recent(ish) regression, but the code has been filtering on even line numbers since at least 2014, which implies that the output of runtime.Stack() has been like this for at least that long (or the test would have failed already earlier).

The `strings.HasPrefix(stack[0], "goroutine ")` check was meant as safeguard against Go versions which don't have that line. Now I am not sure anymore whether that check is necessary. We might as well just remove the first line unconditionally. But better safe than sorry, so I prefer to keep it.

> The PruneStack function has tests. Is it possible to extend those tests, or add some testing elsewhere?

I have extended those tests. The problem with them was that they didn't actually test `PruneStack` on the output of `runtime/debug.Stack()`, so changes in that output weren't detected. I added that.

One thing I noticed is that full stack traces now end with:
```
  testing.tRunner(0xc00045a800, 0x1ad27b8)
  	/nvme/gopath/go/src/testing/testing.go:865 +0xc0
  created by testing.(*T).Run
  	/nvme/gopath/go/src/testing/testing.go:916 +0x35a
```

`/nvme/gopath/` is my `$GOPATH`. Is that something that used to be filtered out by the `/pkg/testing` match in `PruneStack`? It looks a bit like it.

I would prefer to have `PruneStack` filter out those entries, which implies extending the regex and the test to ensure that it does.

blgm:
Thank you for your careful analysis @pohly

pohly:
@blgm: what should be done about `go/src/testing` pruning? Do you agree that this should be filtered out?

blgm:
@pohly, in an ideal world I agree that it should be pruned too as it does not add value to the message. If you want to submit another PR to do that since you have context, I'd happily review it. But I think it's a very minor thing, so I was quite happy to merge this PR without that.

pohly:
> But I think it's a very minor thing, so I was quite happy to merge this PR without that.

True. Let's leave it like it is now. The path matching is already a bit brittle, I don't want to make that worse. Will there be a release with this fix included soonish?

blgm:
Of course. Here's a new patch release: https://github.com/onsi/ginkgo/releases/tag/v1.10.0

PR #593, opened 2019-07-20
Simplify code

muesli:
- Removed unnecessary returns
- Simpler boolean comparisons
- No need to specify capacity of 0, which is the default
- Use fmt.Errorf


williammartin:
All LGTM thanks. Did you use some automated tool to find all these?

Issue #320, opened 2017-02-02
Feature request: Set reporter via command line flags


fredr:
It would be great if it was possible to to specify a reporter when running `ginkgo`, so that different reporters can be used in different CI systems etc.

I.e I don't want every developer to have to remember to use `RunSpecsWithDefaultAndCustomReporters` for the tests to be picked up by CI, and I would like to not have to change every TestSuite whenever I need to add a new output format.

msabramo:
👍 As I want to add junit reporting and would like to avoid having to modify 24 `*_suite_test.go` files.

I'm willing to take a stab at it, but I'd like to get feedback from @onsi before diving in.

msabramo:
Here's my workaround:

```
sed -i -e 's#RunSpecs(\(.*\))$#junitReporter := reporters.NewJUnitReporter("junit.xml")\'$'\n'$'\t''RunSpecsWithDefaultAndCustomReporters\(\1, []Reporter{junitReporter}\)#' $(find . -name '*vendor*' -prune -o -name '*_suite_test.go' -print)

goimports -w $(find . -name '*vendor*' -prune -o -name '*_suite_test.go' -print)
```

manosnoam:
Hi, I've added this PR #598 - it is passing on [travis-ci](https://travis-ci.org/onsi/ginkgo/builds/571680500). 
Please ignore the failed builds on my personal fork, probably I need to reset my fork.
@williammartin can you review please ?

manosnoam:
@msabramo, @fredr,
You can try to pull this patch by adding to your go.mod:
`replace github.com/onsi/ginkgo => github.com/manosnoam/ginkgo v0.0.0-20190902135631-1995eead7451`

And then run:

`$ go test -v <path of packages to test> -ginkgo.reportFile junit.xml`

This is an output example:
```

=== RUN   TestRoute
 Running Suite: Route Suite
 ==========================
 Random Seed: 1567587744
 Will run 4 of 4 specs
 
 ••••
 JUnit path was configured: junit.xml
 
 JUnit report was created: /home/nmanos/go/src/github.com/submariner-io/submariner/pkg/routeagent/controllers/route/junit.xml
 
 Ran 4 of 4 Specs in 0.000 seconds
 SUCCESS! -- 4 Passed | 0 Failed | 0 Pending | 0 Skipped
 --- PASS: TestRoute (0.00s)
 PASS
 
=== RUN   TestUtil
 Running Suite: Util Suite
 =========================
[...]

PR #601, opened 2019-09-02
A new CLI option: -ginkgo.reportFile <file path>

manosnoam:
If specified, it will change the Junit output file path,
that was pre-configured in ginkgo test suite.
It will create parent directories, if not already exist.

Use this option in order to avoid modifying and building
the suite test files, just to specify new report file path.

For example, when running with:
`--reportFile ../../a_new_dir_1/a_new_dir_2/junit.xml`

The junit.xml report will be created under:
`<current path>../../a_new_dir_1/a_new_dir_2/`

Note the changes in RunSpecs() of ginkgo_dsl.go:
It will create a new Junit file, even if Junit report
was not predefined in the test suites. I.e.:
When using this CLI option, there's no longer need to call:
```
reporterList := []ginkgo.Reporter{}
reporterList = append(reporterList, reporters.NewJUnitReporter(junitPath))
ginkgo.RunSpecsWithDefaultAndCustomReporters(t, "Test suite", reporterList)
```

[Fixes onsi#320]

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -53,6 +53,7 @@ type DefaultReporterConfigType struct {
 	Verbose           bool
 	FullTrace         bool
 	ReportPassed      bool
+	ReportFile        string
 }
 
 var DefaultReporterConfig = DefaultReporterConfigType{}
@@ -100,6 +101,8 @@ func Flags(flagSet *flag.FlagSet, prefix string, includeParallelFlags bool) {
 	flagSet.BoolVar(&(DefaultReporterConfig.Succinct), prefix+"succinct", false, "If set, default reporter prints out a very succinct report")
 	flagSet.BoolVar(&(DefaultReporterConfig.FullTrace), prefix+"trace", false, "If set, default reporter prints out the full stack trace when a failure occurs")
 	flagSet.BoolVar(&(DefaultReporterConfig.ReportPassed), prefix+"reportPassed", false, "If set, default reporter prints out captured output of passed tests.")
+	flagSet.StringVar(&(DefaultReporterConfig.ReportFile), prefix+"reportFile", "", "Override the default reporter output file path.")
+
 }
 
 func BuildFlagArgs(prefix string, ginkgo GinkgoConfigType, reporter DefaultReporterConfigType) []string {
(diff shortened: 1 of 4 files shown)


manosnoam:
@blgm can you take a look at this feature enhancement, that was requested by several users on #320 ?

manosnoam:
@blgm Thanks so much for quick response!
I'll add a DOC PR for this change as well.
master is good to go for others to try out this feature, and later I think a new release would be great.

manosnoam:
> @blgm Thanks so much for quick response!
> I've added a DOC PR #613 for this change as well.
> `master` is good to go for others to try out this feature, and later I think a new release would be great.

Hi @blgm Can you add it to release?
It's out on master for a month now, and Ginkgo docs show this option already. Thanks!

blgm:
Thank you for the reminder @manosnoam. New release here: https://github.com/onsi/ginkgo/releases/tag/v1.10.3

segevfiner:
Can this be used when running in parallel? It doesn't look like it as it will set the report file name for all nodes to be the same?

manosnoam:
Hi @segevfiner,
I did run it on parallel packages, but in order to create new report for each package without overriding, the file path must be relative, not absolute, so the output file will be created under each package directory.

segevfiner:
There is also a mode that runs the same package in parallel, I think that is `-p` or `-nodes`.

manosnoam:
In that case, what happens without the use of --reportFile (specifying the
file inside the code instead) - won't it override the file?

I suggest to open an issue to deal with such scenario, if it happens, which
requires to rename existing report, delete it, or just show an error.

segevfiner:
The code looks like it clobbers it so you get only one of the report files from one of the processes. I haven't yet tried it. The documentation shows you need to do something like this:
```go
junitReporter := reporters.NewJUnitReporter(fmt.Sprintf("junit_%d.xml", config.GinkgoConfig.ParallelNode))
```

manosnoam:
Good catch. In that case, can you open an issue/PR to fix and test it?
Seems like a minor fix, but more integration tests of parallel executions are required as well.

PR #607, opened 2019-09-16
Set go_import_path in travis.yml to allow internal packages in forks

manosnoam:
See Travis-CI docs for more details:
https://docs.travis-ci.com/user/languages/go/#go-import-path

[Fixes onsi#599]


.travis.yml
```diff
@@ -5,6 +5,8 @@ go:
   - 1.12.x
   - tip
 
+go_import_path: github.com/onsi/ginkgo
```
blgm: In the future someone might remove this line saying that it's unnecessary, which is true for the original. I wonder whether it's worth adding a comment to say that this has been added to the original as a convenience for forks?

manosnoam: Yes, good idea.
Another option to allow forked repo to pass travis-ci, is to add this change to the forkrd repo only, but that will not be permanent, once the fork is reset to match upstream.

.travis.yml
```diff
@@ -5,6 +5,9 @@ go:
   - 1.12.x
   - tip
 
+# allow internal package imports, necessary for forked repositories
```
manosnoam: @blgm Is that fine comment ?

Issue #608, opened 2019-09-19
extension/table driven test may be slow


avamingli:
When using table driven extension, the test may be slow if there are a lot of params and a lot of table entries. Something like that:

```
table.DescribeTable("just a example", func(params ...interface{}) {
    // do something
    },
        table.Entry("description 1", "15", 1, 3, 4, 5, 6, 7, 8, 9),
	table.Entry("description 2", "6", 1, 3, 4, 5, 6, 7, 8, 9),
	table.Entry("description 3", "101", 0, 2, 4, 5, 6, 7, 8, 9),
         // other entries
	table.Entry("description n", "5", 10, 2, 4, 5, 6, 7, 8, 9),
	table.Entry("description n+1", "0", 3, 1, 4, 5, 6, 7, 8, 9),
)

```

avamingli:
Fixed in pr https://github.com/onsi/ginkgo/pull/609

PR #609, opened 2019-09-19
speed up table entry generateIt()

avamingli:
speed up the func generateIt() when using table driven tests.
Fix issue(https://github.com/onsi/ginkgo/issues/608)


blgm:
Thank you @avamingli!

PR #626, opened 2019-11-15
teamcity reporter: output location of test failure

### LINKED ISSUE #623: Teamcity Reporter does not include location of failure
Unlike the JUnit reporter, the TeamCity reporter does not include the location of failure. It includes only the location of the test (i.e. its first line):
```
##teamcity[testFailed name='Test group Second test' message='tests.go:33' details='Unexpected error:|n    <*errors.errorString || 0xc0001b8000>: {|n        s: "failed on purpose",|n    }|n    failed on purpose|noccurred|n']
```
When reviewing a failed test in TeamCity, it's helpful to know the location of the failure.

dlipovetsky:
Fixes #623

### DIFF
--- a/reporters/teamcity_reporter.go
+++ b/reporters/teamcity_reporter.go
@@ -50,8 +50,8 @@ func (reporter *TeamCityReporter) handleSetupSummary(name string, setupSummary *
 	if setupSummary.State != types.SpecStatePassed {
 		testName := escape(name)
 		fmt.Fprintf(reporter.writer, "%s[testStarted name='%s']\n", messageId, testName)
-		message := escape(setupSummary.Failure.ComponentCodeLocation.String())
-		details := escape(setupSummary.Failure.Message)
+		message := reporter.failureMessage(setupSummary.Failure)
+		details := reporter.failureDetails(setupSummary.Failure)
 		fmt.Fprintf(reporter.writer, "%s[testFailed name='%s' message='%s' details='%s']\n", messageId, testName, message, details)
 		durationInMilliseconds := setupSummary.RunTime.Seconds() * 1000
 		fmt.Fprintf(reporter.writer, "%s[testFinished name='%s' duration='%v']\n", messageId, testName, durationInMilliseconds)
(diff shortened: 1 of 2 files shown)


blgm:
Thank you @dlipovetsky. I don't have much context on the teamcity reporter, and it has only been changed once in the past few years - by @manosnoam. So I'll leave this a few days in case @manosnoam has any feedback, since sometimes changes to formats can sometimes break other teams.

manosnoam:
Hi @dlipovetsky,
What's the difference between `failure.ComponentCodeLocation.String()` and `failure.Location.String()`

Can you give an output (or screenshot) example from TeamCity ?

dlipovetsky:
Hi @manosnoam!

Thanks for reviewing, and for implementing the TeamCity reporter :smile: 

> Hi @dlipovetsky,
> What's the difference between `failure.ComponentCodeLocation.String()` and `failure.Location.String()` ?

`ComponentCodeLocation` is the line of code that defines the Ginkgo "Component" (Ginko's internal name for what its docs refer to as blocks and specs, i.e., Describe/Context/It/BeforeSuite, etc). See [here](https://github.com/onsi/ginkgo/blob/0ecbc58698389c4a38c1b07e092f686286a73013/types/types.go#L155) for some details; unfortunately, there aren't any comments.

`Location` is the line of code where the test failed, e..g, the line where an assertion failed. When a test fails,  `Location`  is an important piece of information. Without it, you need to search where precisely the test failed.

For comparison, the JUnit reporter [outputs both locations](https://github.com/onsi/ginkgo/blob/v1.10.3/reporters/junit_reporter.go#L90) for details. when a test fails.

> Can you also give an output (or screenshot) example from TeamCity ?

I'll try to get one that doesn't leak any internal information :smile: .

dlipovetsky:
> So I'll leave this a few days in case @manosnoam has any feedback, since sometimes changes to formats can sometimes break other teams.

Thanks for the review @blgm! You make a good point.

The TeamCity message contains two attributes, `message` and `details`. This PR adds a single line of information to `details`.

For example:

Before:
```
[testFailed name="some test" message=''/path/to/test.go:100 details="unlucky"]
```
After
```
[testFailed name="some test" message=''/path/to/test.go:100 details="unlucky|n/path/to/test.go:20"]
```

I only expect TeamCity itself to parse the service messages. I don't believe TeamCity parses the contents of `details`  at all. While TeamCity collects test statistics based on the test name; the message and details are not relevant.

dlipovetsky:
Rebased.

dlipovetsky:
@manosnoam Your thoughts?

manosnoam:
Looks good to me. 

A side note:
The location of test failure in code is very helpful, but it is not always clear in Ginkgo output/stacktrace.

For example, it has happened to me that an assertion failed inside an `It` block:
https://github.com/onsi/ginkgo/blob/3cfa02ded2d565a682638855d5f40dd9e00b6394/reporters/junit_reporter_test.go#L192

But then the stacktrace did not refer to that exact line of assertion, but only to the first line of the block:
https://github.com/onsi/ginkgo/blob/3cfa02ded2d565a682638855d5f40dd9e00b6394/reporters/junit_reporter_test.go#L183

In this case it is unknown what assertion actually broke the test...
This of-course should be opened as a different issue to fix.

blgm:
@dlipovetsky thank you for this contribution!

@manosnoam thank you for reviewing!

PR #635, opened 2019-12-31
feat: table tests should point to test code

### LINKED ISSUE #515: Correct code location for table driven tests
For table driven tests, the code location is always showing https://github.com/onsi/ginkgo/blob/master/extensions/table/table_entry.go#L46 which is not correct. This should be a useful feature for some testers.

Is it possible to call PutItNode directly with correct call depth like this? https://github.com/onsi/ginkgo/blob/master/ginkgo_dsl.go#L375

sethp-nr:
Previously, ginkgo would identify the code locations of the failing
Entry or DescribeTable as part of the
ginkgo/extensions/table/table_entry.go or
ginkgo/extensions/table/table.go files, respectively.

When there are a large number of table entries, it is more desirable to
provide the line numbers of the table and table entry directly to
facilitate jumping to the relevant section of the test code.

Fixes #515

### DIFF
--- a/extensions/table/table_entry.go
+++ b/extensions/table/table_entry.go
@@ -4,6 +4,8 @@ import (
 	"reflect"
 
 	"github.com/onsi/ginkgo"
+	"github.com/onsi/ginkgo/internal/codelocation"
+	"github.com/onsi/ginkgo/types"
 )
 
 /*
@@ -12,13 +14,15 @@ TableEntry represents an entry in a table test.  You generally use the `Entry` c
 type TableEntry struct {
 	Description string
 	Parameters  []interface{}
-	Pending     bool
-	Focused     bool
+	Flag        types.FlagType
+
+	// TODO: what if this isn't set?
+	codeLocation types.CodeLocation
 }
 
 func (t TableEntry) generateIt(itBody reflect.Value) {
-	if t.Pending {
-		ginkgo.PIt(t.Description)
+	if t.Flag == types.FlagTypePending {
+		ginkgo.ExplicitItNode(t.Description, func() {}, types.FlagTypePending, t.codeLocation, 0)
 		return
 	}
 
@@ -33,15 +37,9 @@ func (t TableEntry) generateIt(itBody reflect.Value) {
 		}
 	}
 
-	body := func() {
+	ginkgo.ExplicitItNode(t.Description, func() {
 		itBody.Call(values)
-	}
-
-	if t.Focused {
-		ginkgo.FIt(t.Description, body)
-	} else {
-		ginkgo.It(t.Description, body)
-	}
+	}, t.Flag, t.codeLocation, 0)
 }
 
 /*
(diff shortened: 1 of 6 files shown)


ansd:
Hey @sethp-nr, 

Thanks a lot for your PR.
@dpb587-pivotal and I had a look at it last week.
We like it, the changes are good and working, in particular the integration tests!
However, there are a couple of points which I'm going to provide in-line review.
Inspired by your PR, I created #666.
I would be happy about your feedback / thoughts.

ginkgo_dsl.go
```diff
@@ -615,6 +615,22 @@ func AfterEach(body interface{}, timeout ...float64) bool {
...
+}
+
+//ExplicitContainerNode adds a fully-specified Container node to the global context.  This function is
+//largely only useful for meta-programming or adding on to the Ginkgo DSL, when writing normal test
+//cases it's best to use one of the others.
+func ExplicitContainerNode(text string, body func(), flag types.FlagType, codeLocation types.CodeLocation) bool {
+	globalSuite.PushContainerNode(text, body, flag, codeLocation)
+	return true
+}
+
```
ansd, line 633: Also exporting these two functions works, it's not ideal. Every user which imports `github.com/onsi/ginkgo` will have access to these functions although these are only internal helper functions.

integration/run_test.go
```diff
@@ -353,6 +354,41 @@ var _ = Describe("Running Specs", func() {
...
+
+			// The entry
+			Ω(output).Should(
+				ContainSubstring("failing_table_tests/failing_table_tests_test.go:14"),
+				"point to the failing entry")
+			Ω(output).ShouldNot(MatchRegexp(`github.com/onsi/ginkgo/extensions/table/table_entry.go:\[\d+\]`))
+
+			Ω(output).Should(
+				ContainSubstring("failing_table_tests/failing_table_tests_test.go:12"),
+				"point to the failing assertion")
```
ansd, line 388: It's cool that you provide the optional test description given the complexity of the test.

extensions/table/table_entry.go
```diff
@@ -12,13 +14,15 @@ TableEntry represents an entry in a table test.  You generally use the `Entry` c
 type TableEntry struct {
 	Description string
 	Parameters  []interface{}
-	Pending     bool
-	Focused     bool
+	Flag        types.FlagType
```
ansd, line 17: I also first thought it's a good idea to use the single `Flag` field instead of `Pending` and `Focused`. I'm not sure since I haven't written the original code but it looks like the `Pending` and `Focused` fields are there on purpose since they are exported to the user?
Also, if we use do this change, we will break all users who construct a `TableEntry` directly (rather than using the `(F/P/X)Entry` constructors). Looking at the go docs for [TableEntry](https://github.com/onsi/ginkgo/blob/master/extensions/table/table_entry.go#L12), they state

> TableEntry represents an entry in a table test. You **generally** use the `Entry` constructor.

Although I've not seen people constructing the `TableEntry` directly, some people might do in which case we would break them with these changes.

extensions/table/table_entry.go
```diff
@@ -12,13 +14,15 @@ TableEntry represents an entry in a table test.  You generally use the `Entry` c
...
+
+	// TODO: what if this isn't set?
```
ansd, line 19: Good point. I don't have a clear answer. In #666 I just print the surrounding `DescribeTable` whenever it's not set.

extensions/table/table_entry.go
```diff
@@ -33,15 +37,9 @@ func (t TableEntry) generateIt(itBody reflect.Value) {
...
+	ginkgo.ExplicitItNode(t.Description, func() {
 		itBody.Call(values)
-	}
-
-	if t.Focused {
-		ginkgo.FIt(t.Description, body)
-	} else {
-		ginkgo.It(t.Description, body)
-	}
+	}, t.Flag, t.codeLocation, 0)
```
ansd, line 42: The last `0` argument does not look correct?
You pass it further to `ExplicitItNode` where it becomes the timeout for `PushItNode` which would result in a timeout of 0 seconds [here](https://github.com/onsi/ginkgo/blob/1ea49cf36e84a951dd339c6da927162766ac81c4/ginkgo_dsl.go#L622) rather than the default timeout of 1 second defined [here](https://github.com/onsi/ginkgo/blob/1ea49cf36e84a951dd339c6da927162766ac81c4/ginkgo_dsl.go#L620).

sethp-nr:
That's all great feedback, thanks! It looks like your PR #666 captured all the valuable parts to me, let's close this in favor of that one :)

PR #638, opened 2020-01-14
Use Dup2 from golang.org/x/sys/unix

tklauser:
x/sys/unix provides Dup2 on all unix GOOS/GOARCH variants. The ones for
e.g. linux/arm64 or linux/riscv64 just wrap Dup3 as this package
currently does with syscallDup. Thus, use Dup2 from x/sys/unix to
simplify the implementation in this package.


blgm:
Thank you @tklauser!

PR #642, opened 2020-02-07
skip race detector test on unsupported platform

zhsj:


blgm:
Thank you @zhsj! It's a shame that there isn't a function in the standard library that can be used. I think this is a pragmatic solution.

PR #645, opened 2020-02-14
check if a spec is run before returning SpecSummary

p0lyn0mial:
This pull prevents a null pointer exception by checking if a spec is run before returning its summary. 

The issue might occur when someone simply executes `ginkgo.CurrentGinkgoTestDescription()`


blgm:
Looks good.  Thank you @p0lyn0mial!

PR #663, opened 2020-04-11
fix typo succesful -> successful

the-gigi:
Just a little typo across multiple files.

See: 

https://whichiscorrect.com/successful-or-succesful/

and

https://wikidiff.com/succesful/successful


ansd:
Thanks @the-gigi!

PR #686, opened 2020-06-06
Fix integration tests to stop require GOPATH

alex-slynko:
This still requires gomega and some version of ginkgo in GOPATH.


blgm:
Thank you @alex-slynko

Issue #688, opened 2020-06-08
Dumping table rows parameter in the report


fedepaol:
Hi there, we have a matrix - generated suite of tests, which end up as different table rows of a table.
What we would like to do is to have the different parameters dumped in the test output, shall it be the one on standard output or the junit report.

What is the suggested way to have such behaviour? Is dumping the parameters in the entry description the only way? 
Would it make sense to (optionally) dump out the parameters in case of table tests? After all, the parameters are part of what's being tested in that particular row. Is this something you would accept as a change?

blgm:
Hi @fedepaol. I'm assuming that you're talking about `DescribeTable()`. Here are some ways that people commonly log more details about their test:

- [`GinkgoWriter`](http://onsi.github.io/ginkgo/#logging-output)
- [`By()`](http://onsi.github.io/ginkgo/#documenting-complex-its-by)
- If you are using Gomega, you can [annotate assertions](http://onsi.github.io/gomega/#annotating-assertions)

I can see the appeal of always logging the parameters in a `DescribeTable()`, but I also see some cases where it might be hard to format them. For instance if testing subtle differences in large structs it could be hard to see what's different. If an input parameter is a function then what would you log? My initial reaction is that Ginkgo should give test writers the ability to easily log whatever they like, and should not make assumptions about what a test will want to log. However I am open to being persuaded otherwise.

fedepaol:
Thanks @blgm for replying. Yes, I am talking about `DescribeTable` and the various `Entry` parameters.

Our need is to be able to dump the parameters in some way, because they are part of the single row description. We may even want to focus / skip some of them.
We started using GinkgoWriter, but the solution we came with felt a bit clunky because of the mandatory lines we needed to add to the tests (plus, by doing that we can't use the parameters for skipping / focusing only on some tests), hence the request for discussion.

On one side, the points you are making make a lot of sense (func parameters, structs to big to log), on the other hand it still feels that something is missing because a given parameter is still describing the specific test that entry is referring to.
When a test fail, I would like to know which parameters it failed with, and there's currently no "native way" if not injecting them in the entry description when I create the entries.
But if you think there can't be an (even optional) way to change that, I'll live with it :-)

Thanks for the help

blgm:
I think it's always worth having a discussion.

[...]

PR #689, opened 2020-06-08
Add a version of table.Entry that allows dumping the entry parameters.

fedepaol:
Sometimes, it's useful to have the parameters of a given table entry as part of the description of the It counterpart.
To achieve that, we add a new family of Entry constructor that accept a function instead of the description as a string.
That function needs to be implemented by the client of the library, and it's called with the list of arguments and returns a string that will be used as description of the given entry.

This was discussed in https://github.com/onsi/ginkgo/issues/688

I am not totally happy with the naming, but I guess keeping it short-ish has a value here.

### DIFF
--- a/extensions/table/table.go
+++ b/extensions/table/table.go
@@ -42,6 +42,25 @@ Under the hood, `DescribeTable` simply generates a new Ginkgo `Describe`.  Each 
 It's important to understand that the `Describe`s and `It`s are generated at evaluation time (i.e. when Ginkgo constructs the tree of tests and before the tests run).
 
 Individual Entries can be focused (with FEntry) or marked pending (with PEntry or XEntry).  In addition, the entire table can be focused or marked pending with FDescribeTable and PDescribeTable/XDescribeTable.
+
+A description function can be passed to Entry in place of the description. The function is then fed with the entry parameters to generate the description of the It corresponding to that particular Entry.
+
+For example:
+
+	describe := func(desc string) func(int, int, bool) string {
+		return func(x, y int, expected bool) string {
+			return fmt.Sprintf("%s x=%d y=%d expected:%t", desc, x, y, expected)
+		}
+	}
+
+	DescribeTable("a simple table",
+		func(x int, y int, expected bool) {
+			Ω(x > y).Should(Equal(expected))
+		},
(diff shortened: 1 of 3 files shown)


onsi:
Thanks @fedepaol - this is in the right direction but now I'm realizing why we might actually want to use some magic.  The current implementation is not symmetric with the existing implementation of `DescribeTable`.  Specifically, the function passed to `DescribeTable` has a signature that describes the inputs of the test (e.g. from the test you've added: `func(x int, y int, expected bool)` however the function passed to `DEntry` will be forced to have a less descriptive signature: `func(params []interface{}) string` that will force the user to refer to the parameters by by index instead of by name.

So I think we actually want `DEntry` to take `interface{}` (just like `DescribeTable` does) and allow the user to pass a function that more closely parallels their `DescribeTable` function.

At that point one could argue that we don't need `DEntry` and can just go with having `Entry` do the right thing depending on the type of input it has... but I don't feel too strongly about that.

WDYT?

fedepaol:
The new version is a bit more complex on code, but I do agree, it looks a bit cleaner on client site.
Please take a look.

onsi:
This looks great, @fedepaol - I think all it needs now is to be documented (inline in the comments, but also on the `gh-pages` branch here: https://github.com/onsi/ginkgo/blob/gh-pages/index.md#table-driven-tests )

If you're up for adding some documentation let me know.  Otherwise, I'll pull this in and document it myself.

fedepaol:
> This looks great, @fedepaol - I think all it needs now is to be documented (inline in the comments, but also on the `gh-pages` branch here: https://github.com/onsi/ginkgo/blob/gh-pages/index.md#table-driven-tests )
> 
> If you're up for adding some documentation let me know. Otherwise, I'll pull this in and document it myself.

Sure, I'll do! TBH, it was pretty late yesterday and I wasn't able to find the docs :P

fedepaol:
@onsi done! Feel free to correct it.

onsi:
Looks great!  I’ll merge these in when i get to my desk.

fedepaol:
Thanks! When do you expect this to be available as a tag?

onsi:
I expect in the next couple of days or so

PR #692, opened 2020-06-16
Add `ResetGlobal` function

nogates:
Hello! First of all, thank you for this amazing project!! ❤️ 

We are currently using Ginkgo to dynamically generate some tests (depending on some configuration) and it works great. However, we are having some troubles to test different scenarios where different `describe` or `context` groups will be created. Since ginkgo stores everything under one global `Suite` struct and there is currently no way to reset that variable, it's very hard for us to test this logic (we are testing the tester, I know! :) 

In this PR, I am adding a `ResetGlobals` function that is exposed as part of the ginkgo DSL, which just resets the globals variables to the original state, and I wanted to get some feedback before spending more time adding tests (I was going to open an Issue first, but given that little amount of code that's needed to show how this could be implemented, I decided to went ahead and open a PR instead, I hope this is ok!)

What do you think about this? Do you think we should do this in a different way? 

Thank you!

### DIFF
--- a/extensions/globals/globals.go
+++ b/extensions/globals/globals.go
@@ -1,0 +1,26 @@
+// Package `globals` provides an interface to alter the global state of ginkgo suite.
+//
+// ginkgo currently registers a few singleton global vars that hold all the
+// test blocks and failure management. These vars are global per package, which means
+// that only one Suite definition can coexist in one package.
+//
+// However, there can be some use cases where applications using ginkgo may want to
+// have a bit more control about this. For instance, a package may be using ginkgo
+// to dynamically generate different tests and groups depending on some configuration.
+// In this particular case, if the application wants to test how these different groups
+// are generated, they will need access to change these global variables, so they
+// can re-generate this global state, and ensure that different configuration generate
+// indeed different tests.
+//
+// Note that this package is not intended to be used as part of normal ginkgo setups, and
+// usually, you will never need to worry about the global state of ginkgo
+package globals
+
+import "github.com/onsi/ginkgo/internal/global"
+
+// Reset calls `global.InitializeGlobals()` which will basically create a new instance
+// of Suite, and therefore, will effectively reset the global variables to the init state.
+// This will effectively remove all groups, tests and blocks that were added to the Suite.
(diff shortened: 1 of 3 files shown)


onsi:
Thanks @nogates - and cool use case you have going :)

I can't think of an alternative approach.  My only hesitation is including `ResetGlobals` in the top-level Ginkgo DSL.  I think, as an alternative, this could go in the `extensions` folder as a new standalone package that simply exposes `ResetGlobals`.    We should think about what to call that package - any suggestions?

nogates:
Hey @onsi ! Thanks for the quick reply :) 

> I can't think of an alternative approach. My only hesitation is including `ResetGlobals` in the top-level Ginkgo DSL. I think, as an alternative, this could go in the `extensions` folder as a new standalone package that simply exposes `ResetGlobals`.

Yeah, that could work! Happy to move it there!

> We should think about what to call that package - any suggestions?

well, I think `globals` (as in `globals.Reset()`), is the most descriptive one... I can't think in any other name that would be more accurate and less generic 😬

onsi:
> well, I think globals (as in globals.Reset()), is the most descriptive one... I can't think in any other name that would be more accurate and less generic 😬

Fair enough - would you be up for adding documentation explaining the original use-case?  Within the context of an actual test suite running `globals.Reset()` would be no bueno - so we want to make sure folks understand that this is intended for meta testing.

nogates:
> Fair enough - would you be up for adding documentation explaining the original use-case? Within the context of an actual test suite running `globals.Reset()` would be no bueno - so we want to make sure folks understand that this is intended for meta testing.

Definitely! I forced pushed the branch and moved the `ResetGlobals` to the `extensions/globals#Reset` package as discussed! 

Also, I added a long description to the beginning of that package explaining what is the purpose of it, and I did emphasize this functionality is not intended for normal ginkgo setups.

Let me know if all of this makes sense! Thanks a lot for the time reviewing this @onsi ! 🙇

onsi:
Great!  lgtm!

nogates:
Thank you!

PR #721, opened 2020-09-23
correct handling windows backslash in import path

bakito:
When using the generate command in windows with go mod enabled modules, the import in the generated test file is generated with windows backslash instead of forward slash.

This PR corrects and replaces the backslash into a forward slash to make the import path valid.


onsi:
lgtm thanks!

PR #726, opened 2020-10-01
Adding support for PPC64LE

ezeeyahoo:
Hi,

I have added PPC64LE to arch: to enable support for it, request you to review it.

Thanks


williammartin:
It doesn't seem like this is really wanted, and we don't use Travis anymore, so I'm going to close this and if anyone complains we can revisit.

Issue #735, opened 2020-11-06
[feature] [cli] ability to specify multiple -focus and -skip flags


kolyshkin:
I would like to have an ability to specify multiple `-focus` and `-skip` flags. Here's why.

Currently, I have something like this in one of the CI files:
> -skip=\[Slow\]|\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|PersistentVolumes|\[HPA\]|should.support.building.a.client.with.a.CSR|should.propagate.mounts.to.the.host|for.NodePort.service|type.clusterIP|unready.pods|ExternalName.services|Guestbook.application|in-cluster.config|Pods.should.support.pod.readiness.gates|\[sig-storage\].In-tree.Volumes.\[Driver:.local\]|\[sig-storage\].CSI.Volumes.CSI.Topology.test.using.GCE.PD.driver|\[sig\-storage\]\sCSI\sVolumes\s\[Driver\:\scsi\-hostpath\]\s\[Testpattern\:\sDynamic\sPV\s\(block\svolmode\)\]\svolumes\sshould\sstore\sdata|\[sig\-storage\]\sCSI\sVolumes\s\[Driver\:\scsi\-hostpath\]\s\[Testpattern\:\sDynamic\sPV\s\(block\svolmode\)\]\sprovisioning\sshould\sprovision\sstorage\swith\spvc\sdata\ssource|\[sig-network\]\sServices\sshould\sbe\sable\sto\spreserve\sUDP\straffic\swhen\sserver\spod\scycles\sfor\sa\sNodePort\sservice|Events.should.be.sent.by.kubelets.and.the.scheduler.about.pods.scheduling.and.running|\[sig\-apps\].DisruptionController.should.block.an.eviction.until.the.PDB.is.updated.to.allow.it|\[sig\-instrumentation\].MetricsGrabber

and yes, it's a mess. This, though, can be made much less messy by having something like this instead:

```
       -skip=\[Slow\]
       -skip=\[Serial\]
       -skip=\[Disruptive\]
       -skip=\[Flaky\]
       -skip=\[Feature:.+\]
       -skip=PersistentVolumes
       -skip=\[HPA\]
       -skip=should.support.building.a.client.with.a.CSR
       -skip=should.propagate.mounts.to.the.host
       -skip=for.NodePort.service
... and so on ...
```

Implementation-wise it seems quite doable, I think I'll have a PR soon. Hope that would be usable.

[...]

PR #736, opened 2020-11-06
cli: allow multiple -focus and -skip flags

kolyshkin:
As an example, now instead of

	... --skip one|two|three|four ...

one can specify

	... \
	--skip one \
	--skip two \
	--skip three \
	--skip four \
	...

which helps readability and maintainability a lot in case there are many
tests to skip.

Fixes https://github.com/onsi/ginkgo/issues/735 (check it for more real-world example).

### DIFF
--- a/internal/suite/suite_test.go
+++ b/internal/suite/suite_test.go
@@ -40,7 +40,7 @@ var _ = Describe("Suite", func() {
 			runOrder             []string
 			randomizeAllSpecs    bool
 			randomSeed           int64
-			focusString          string
+			focusStrings         []string
 			parallelNode         int
 			parallelTotal        int
 			runResult            bool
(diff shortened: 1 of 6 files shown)


kolyshkin:
@onsi PTAL

onsi:
whoops - sorry I missed this PR @kolyshkin .  I'll take a look sometime in the next few days and merge it in.

onsi:
sorry for the delay - this looks good!

internal/suite/suite_test.go
```diff
@@ -186,7 +186,7 @@ var _ = Describe("Suite", func() {
 
 		Context("when provided with a filter", func() {
 			BeforeEach(func() {
-				focusString = `inner|\d`
```
pohly, line 189: This is a breaking change. Users who relied on `-ginkgo.skip` overwriting the old value will be surprised... I certainly was and it took me a while to figure out why some tests no longer ran.

The reason was this default:
```
	// Skip slow or distruptive tests by default.
	flag.Set("ginkgo.skip", `\[Slow|Disruptive\]`)
	flag.Parse()
```

We overwrite that default in our Makefile, but because of this change, the skip string above remained in effect.

moonsn: +1
maybe need keep this test, keep no breaking change.

PR #751, opened 2020-12-11
Add _internal to filename of tests created with internal flag

### LINKED ISSUE #750: Generate internal should use add "_internal" to the test name
When running the `ginkgo generate -internal` it creates a test file with the same name as without the `-internal` flag.
Most of the time I want to generate white box and blackbox tests.

It would be handy if adding the `-internal` flag would add `_internal` to the generated file name:

```
$ something_internal_tests.go
```

From looking around, it looks like this is a convention for internal tests.

kevinkirkup:
Fixes #750


onsi:
also looks good - though it seems the entire `-internal` flag is not covered by the integration tests.  I'll fix that after I pull this PR in and your other one as well.

PR #757, opened 2020-12-31
Document the `ginkgo outline` command

dlipovetsky:
The `ginkgo outline` command was added in https://github.com/onsi/ginkgo/pull/754.

PR #759, opened 2021-01-10
Add sidebar link to section on `ginkgo outline` subcommand

dlipovetsky:
The link was missing from the original document PR (#757).


dlipovetsky:
I added the entry by hand, because it _seemed_ like `ginkgo_sidebar.html` is not generated, but LMK ifthat's incorrect.

onsi:
yep that's fine - no auto-get magic here :)

PR #680, opened 2020-05-28
solve the undefined: unix.Dup2 compile error on mips64le

XiaodongLoong:
error in detail:
../internal/remote/output_interceptor_unix.go:41:2: undefined: unix.Dup2
../internal/remote/output_interceptor_unix.go:42:2: undefined: unix.Dup2

there is Dup2 syscall only on amd64, other arch is Dup3 instead.

Signed-off-by: Xiaodong Liu <liuxiaodong@loongson.cn>

PR #761, opened 2021-01-12
internal/remote: use unix.Dup2 instead of wrapping Dup2/Dup3 manually

tklauser:
The latest version of the golang.org/x/sys/unix package already
implements func Dup2 using Dup3 on platforms which dont have the dup2
syscall. Use Dup2 and let x/sys/unix figure out the correct syscall.

This effectivly reverts PR #680. Verified that ginkgo still correctly
builds on linux/mips64le.


onsi:
sweet!  thank you :) this cleans things up nicely.

PR #762, opened 2021-01-17
outline: Do not omit an empty `nodes` field when marshalling to JSON

dlipovetsky:
If an empty 'nodes' field is represented as an empty JSON array, then a client
unmarshalling from JSON does not need to allow for the possibility that the
'nodes' field is undefined. That makes it is easier for clients to operate on
the JSON.

Before:
```json
{
  "name": "It",
  "text": "normal",
  "start": 336,
  "end": 363,
  "spec": true,
  "focused": false,
  "pending": false
}
```

```json
{
  "name": "It",
  "text": "normal",
  "start": 336,
  "end": 363,
  "spec": true,
  "focused": false,
  "pending": false,
  "nodes": []
}
```


dlipovetsky:
I've been working on a vscode extension that calls `ginkgo outline`. It is much simpler for the client to use array semantics for the `nodes` field. For example, the client can always iterate over `nodes`, and assume the iteration will exit immediately for a leaf node, where `nodes` has zero length. Without this change, the client must check whether `nodes` is defined  before iterating over it.

Feedback welcome!

PR #763, opened 2021-01-18
outline: Report absolute offsets of spec/container start and end positions

dlipovetsky:
This fixes an off-by-one error in the offsets reported for the spec/container
start and end positions. The original implementation used the `Pos` value,
which is not an absolute offset, but is relative to the file in the set of
parsed files:

> The Pos value for a given file is a number in the range [base, base+size],
> where base and size are specified when a file is added to the file set.
> -- https://pkg.go.dev/go/token#Pos

Because only one file is parsed, `Pos` is off by one from the absolute offset.

I've added a dedicated test to verify the reported spec/container start and end
positions.


onsi:
hey looks like there are a few merge conflicts (probably because i merged the other pr in too).  mind taking a look @dlipovetsky ?

dlipovetsky:
> hey looks like there are a few merge conflicts (probably because i merged the other pr in too). mind taking a look @dlipovetsky ?

Absolutely. Thanks for heads-up!

Issue #768, opened 2021-02-10
JUnit reporter reports passing tests using an element that is not in the JUnit XML schema


dlipovetsky:
In #586, the JUnit reporter was extended to capture the output of passing tests when the `-reportPassing` flag is used.

For a passing test, the reporter writes:

```xml
<?xml version="1.0" encoding="UTF-8"?>
  <testsuite name="example suite" tests="1" failures="0" errors="0" time="3.026">
      <testcase name="example test" classname="example suite" time="2.000321685">
          <passed>Sample output</passed>
      </testcase>
  </testsuite>
```

However, the `passed` element is **not** in the [JUnit XML schema](https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd). While that alone might a theoretical problem, it is also a practical problem, because some parsers, e.g., the [TeamCity XML Reporting Plugin JUnit parser](https://github.com/JetBrains/teamcity-xml-tests-reporting/tree/master/agent/src/jetbrains/buildServer/xmlReportPlugin/parsers/antJUnit), conform to that schema, ignore the element, and show no output for passing tests.

The fix seems to be to use the `system-out` element instead. I have confirmed that, when using the `system-out` attribute, the TeamCity XML Reporting Plugin Junit parser shows output for a passing test.

```xml
<?xml version="1.0" encoding="UTF-8"?>
  <testsuite name="example suite" tests="1" failures="0" errors="0" time="3.026">
      <testcase name="example test" classname="example suite" time="2.000321685">
          <system-out>Sample output</system-out>
      </testcase>
  </testsuite>
```

(For anyone asking why the `system-err` element is not used: Ginkgo intercepts both stdout and stderr together, and the default reporter writes that output to stdout).

To be clear, I am not a JUnit expert, and I am looking for feedback. I assume the `passed` element was working for @manosnoam, who authored #586.

manosnoam:
Thanks for fixing it @dlipovetsky , indeed  `<system-out>` should be used instead of `<passed>`.
Originally I added it, assuming it's the contra for `<failed>` tag, which is supported.

<passed> is not supported in other apps schemas that parses Junit, for example Polarion. 
(in ReportPortal id did go through though).

PR #769, opened 2021-02-10
reporters/junit: Use `system-out` element instead of `passed`

dlipovetsky:
Proposed fix for #768.

I would like to hear feedback from JUnit reporter users (or JUnit experts, familiar with the XML schema) before we consider merging this. Thanks!


onsi:
LGTM!

PR #775, opened 2021-03-04
Add slim-sprig template functions to bootstrap/generate

### LINKED ISSUE #774: Feature Request: Add functions for test templates
https://github.com/onsi/ginkgo/pull/752 Introduced the ability to use custom templates to generate test files. I's like to add additional template functions for better controlling and enriching the generated templates, for example by using [sprig](https://github.com/Masterminds/sprig)

Willing to PR this if you'd like

HagaiBarel:
Fixes #774


onsi:
thanks @HagaiBarel - this makes sense and looks simple enough.  my only hesitation is that Sprig is pulling in a _lot_ of additional dependencies. is there a similar library that is a little more lightweight?  as you consider your usecase are you needing all of Sprig or simply a subset?

to say a bit more - i know ginkgo is in use in a number of production systems and pipelines where security is a major concern.  growing ginkgo's dependency graph adds that much more to the security burden across those systems.

HagaiBarel:
I can understand and respect that, dependency chains can be an issue.

I'm not familiar with other `go` template libraries as mature and stable as `sprig`, most other template functionalities are full blown template engines (see https://github.com/avelino/awesome-go#template-engines for some examples) which I think is an overkill here and doesn't solve the `deps` problem. There's [gtf](https://github.com/leekchan/gtf) which has a very small footprint but doesn't seem maintained. Maybe [slim-sprig](https://github.com/go-task/slim-sprig) might work?

HagaiBarel:
@onsi any thoughts on this?

blgm:
I can't speak for @onsi, but I was wondering how hard it would be to make this a plugin? It could solve the dependency issue by making it optional, and it could allow folks to use alternative templating languages in the future without having to make a code change. Perhaps an optional flag for the `ginkgo bootstrap` command and a `sprig` command, like:
```
ginkgo bootstrap --templater /usr/bin/sprig
```

This idea might be completely impractical in this case (I don't see a `sprig` command), but I have seen approaches like this allow programs to be decoupled.

onsi:
thanks for the bump @HagaiBarel - I agree, @blgm a more decoupled approach would be ideal.  If that isn't feasible, slim-sprig definitely looks better from a dependencies POV and I'm ok pulling it in.

HagaiBarel:
I think the idea of `plugins` is interesting and there are other parts of `ginkgo` that might benefit from such an approach (e.g. reporters). I don't think it's possible with how `ginkgo` is today, and I'm not sure how to approach such a design but I'm not familiar with the code base to that degree so if you have some ideas around it we can mull them over. I also think it's out of scope for this PR :smiley: 

Anyways @onsi I've switched to `slim-sprig`, let me know if there's anything else needed

PR #798, opened 2021-04-22
Use package path for file location

saschagrunert:
We now trim the package path from the file location to keep the output
reproducible and independent from the build/test system.

### DIFF
--- a/internal/codelocation/code_location.go
+++ b/internal/codelocation/code_location.go
@@ -1,6 +1,7 @@
 package codelocation
 
 import (
+	"go/build"
 	"regexp"
 	"runtime"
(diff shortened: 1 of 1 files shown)


onsi:
hey - thanks for the taking the time to pull together a PR.  It looks like this depends on `GOPATH` which I believe is on its way out of the go language/ecosystem.  Also, I'd prefer to leave the current behavior in place to make it explicitly clear to the user where exactly a failure is occurring unambiguously.

saschagrunert:
Hey, thank you for the feedback. I guess using `-trimpath` is a better alternative then.

PR #830, opened 2021-08-15
Make it possible to build tests for JS/WASM environments.

llamerada-jp:
Hi,
I'm trying to run tests written using ginkgo in JS/WASM environments. With the current code, we get the following error at build time.

```console
$ GOOS=js GOARCH=wasm ginkgo build ./test/wasm/
Compiling wasm...
Failed to compile wasm:

# github.com/onsi/ginkgo
../../ginkgo/ginkgo_dsl.go:261:108: undefined: remote.NewOutputInterceptor
```

By this PR, I could build my tests. I could also run the test by using the following HTML along with it.
I was wondering if I should split the files because the file names and contents would not match, but once I created this PR. I hope anyone can give me some advice.

```HTML
<!doctype html>
<html lang="en">
<head>
  <title>test colonio</title>
  <meta charset="utf-8">
</head>
<body>
  <!-- copy wasm_exec.js from $(shell go env GOROOT)/misc/wasm/wasm_exec.js -->
  <script src="wasm_exec.js" type="text/javascript"></script>
  <script type="text/javascript">
    const go = new Go();
    // rename output file to `test.wasm`
    WebAssembly.instantiateStreaming(fetch("test.wasm"), go.importObject).then((result) => {
      go.run(result.instance);
    }).catch((err) => {
      console.error(err);
    });
  </script>
</body>
</html>
```

The following is an example of the output from the actual test run.

image

Signed-off-by: Yuji Ito <llamerada.jp@gmail.com>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,13 @@ jobs:
     - uses: actions/setup-go@v2
       with:
         go-version: ${{ matrix.version }}
+    - uses: actions/setup-node@v1
+      with:
+        node-version: 14.x
     - uses: actions/checkout@v2
     - run: go mod tidy && git diff --exit-code go.mod go.sum
     - run: go vet ./...
     - run: go run ./ginkgo -r --randomizeAllSpecs --randomizeSuites --race --trace
+    - run: GOOS=js GOARCH=wasm ginkgo build ./...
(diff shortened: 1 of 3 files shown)


onsi:
Cool - I don't know too much about WASM support but will try to take a look at this PR in the next couple of weeks.

llamerada-jp:
I closed this PR because Ginko did major updates and had big changes.

PR #846, opened 2021-10-11
refactor: move from io/ioutil to io and os package

Juneezee:
The `io/ioutil` package has been deprecated in Go 1.16 (See https://golang.org/doc/go1.16#ioutil). Since ginkgo has dropped support for Go 1.15 (81705fc0e78113b1453bd0564ec51918a5a08501), this PR replaces the existing `io/ioutil` functions with their new definitions in `io` and `os` packages.


Juneezee:
@onsi Sorry, the issues should be fixed now. Please take a look again. Thanks!

PR #860, opened 2021-11-12
docs: added TOC for ver2

prashantgupta24:
Added a table of contents to the already awesome readme file


prashantgupta24:
@onsi Do you think you can merge this in? It's really hard to go through the docs without a TOC

onsi:
The new docs site will have a TOC on the sidebar.  Should make it easier to go through and won't need to be manually maintained separately...

onsi:
It looks like this:

image

prashantgupta24:
That's awesome! Will wait for that then.

PR #864, opened 2021-11-22
Missing 'v' in the version to be pulled

marianopinto:
Unknown revision was getting returned after running the `go get github.com/onsi/ginkgo/v2@2.0.0-rc2` command.


onsi:
thanks!

PR #872, opened 2022-01-04
Add error check for invalid/nil parameters to DescribeTable

kralicky:
This PR adds a check for nil Entry parameters to DescribeTable. If a nil entry is (accidentally) given, the reported error message is misleading:
```
Ginkgo detected a panic while constructing the spec tree.
You may be trying to make an assertion in the body of a container node
```

This came up when I wrote a test that looked like this, which caused the above issue and took me a while to find:
```go
var _ = Describe("Test", func() {
  DescribeTable("Test",
    func(..., matchErr interface{}) {
      err := ... // test logic
      if matchErr != nil {
        Expect(err).To(MatchError(matchErr))
      }
    },
    Entry(nil, "test", uuid.New(), []byte("test"), nil),
    Entry(nil, "", uuid.New(), []byte("")), nil,  // <- subtle syntax error!
    Entry(nil, "", uuid.Nil, []byte(""), "error message"),
  )
})
```

The implementation of generateTable panics since it expects reflect.TypeOf(entry) != nil for all entries, and the resulting panic stack trace is hidden from the output.

### DIFF
--- a/table_dsl.go
+++ b/table_dsl.go
@@ -135,8 +135,10 @@ func generateTable(description string, args ...interface{}) {
 		return "Entry: " + strings.Join(out, ", ")
 	}
 
-	for _, arg := range args {
+	for i, arg := range args {
 		switch t := reflect.TypeOf(arg); {
+		case t == nil:
+			exitIfErr(types.GinkgoErrors.IncorrectParameterTypeForTable(i, "nil", cl))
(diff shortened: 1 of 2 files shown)


onsi:
hey, good catch.  Thanks for this!  I'll pull it in once the GitHub actions tests go green.

onsi:
Alrighty, all good.  I'll fold this into the next release in a week or two.

PR #879, opened 2022-01-10
Fix an example in the Spec Parallelization section

yuhaohaoyu:
In the example code in the subsection, Discovering Which Parallel Process a Spec is Running On, the closure variable should be `book`.


onsi:
good catch thanks!

PR #953, opened 2022-03-25
Fix ginkgo import circle

BooleanCat:
The cycle is broken!

By using the latest gomega (with Go 1.18) and updating this go.mod to 1.18, we drop the transitive ginkgo dependencies.


BooleanCat:
I've used an unreleased version of gomega at the moment, will update PR once gomega ships.

onsi:
🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉

onsi:
Just pushed out Gomega 1.19.0!

BooleanCat:
Updated with v1.19.0. That go.sum diff is pretty satisfying 👌🏻

PR #958, opened 2022-04-05
docs: MIGRATING_TO_V2.md update go install cmd

qin-nz:
go: 'go install' requires a version when current directory is not in a module
        Try 'go install github.com/onsi/ginkgo/v2/ginkgo@latest' to install the latest version


onsi:
thanks!

PR #961, opened 2022-04-12
Update Migration docs install command

mudler:
I've come across this multiple time, now I realized the command I was copying from migration was not inline with front docs, which is the one that actually works (here locally I get go mod issues with it if running it on a existing project)


onsi:
sorry for the delay - thanks for the PR!

PR #978, opened 2022-05-15
refactor: replace strings.Replace with strings.ReplaceAll

Juneezee:
Replace `strings.Replace(s, old, new, -1)` with `strings.ReplaceAll(s, old, new)`. `strings.ReplaceAll` is a wrapper function for `strings.Replace`, but `strings.ReplaceAll` is more readable and removes the hardcoded `-1`.


blgm (approved):
We know that some folks use quite old versions of go. The ReplaceAll() function was added in go1.12, which is long enough ago that this change shouldn't cause issues.

blgm:
Thank you @Juneezee

PR #996, opened 2022-06-20
report: print summary if failure exists

Letty5411:
Currently, if the failure equals one, `summarizing` will not print. This behavior brings some trouble when I have a test results analysis tool which relays on the key word `[FAIL]`.


onsi:
I'm closing this as I've implemented it on master and [this](https://github.com/onsi/ginkgo/pull/996/commits/7d02b23591498389b790fa21a4b6be7e94435b59) commit seemed unrelated to the issue at hand.

PR #997, opened 2022-06-25
Fixed a couple of minor typos

anshuman-agarwala:


blgm:
Thank you @anshuman-agarwala!

Issue #873, opened 2022-01-05
Detecting if a suite has failed in `AfterSuite` blocks


austince:
Hi there, I'm currently trying to check if any of the specs in a suite have failed during test cleanup. If so, I'd like to collect some debug information and write it to a file before cleaning up the created resources.

In ginkgo 1, a custom reporter works ok for this, but I know they have been removed in ginkgo 2. I've considered the new `ReportAfterEach`, but these run after the cleanup code and feels like abusing the constructs to put cleanup code in there. Is there another way to detect if a suite has failed before cleaning up? 

Thanks!

onsi:
hey @austince - I think it would help to have a bit more detail about the use case and lifecycle for your specs but some quick thoughts ahead of that.

If the intent is to save off debug information after each failed spec then I'd put that code in an `AfterEach` and use [`CurrentSpecReport()`](https://onsi.github.io/ginkgo/#getting-a-report-for-the-current-spec) to figure out if the spec failed or not.  If it failed - I'd store off the debug information and then cleanup.

One question is _where_ to put the debug information.  You can write it to the filesystem but that could get tricky if you want to run in parallel.  In Ginkgo V2 I'd suggest adding the debug information as a `ReportEntry` via [`AddReportEntry`](https://onsi.github.io/ginkgo/#attaching-data-to-reports).  This information will then end up in the JSON/JUnit output (if you use `--json-report/--junit-report`) and on the console.

If the intent is to save off debug information at the _very end_ of the entire suite then I'd use `ReportAfterSuite` to figure out what failed and then save off the diagnostics.

If what you want is to aggregate debug data for failures as the suite runs and then produce a single coherent report in your own format that includes the debug information I'd combine the two approaches.  I'd use `AddReportEntry` to attach information to each failed spec, then I'd use `ReportAfterSuite` to walk through the specs, find the failed specs, pull and rehydrate the relevant `ReportEntries` and generate your own custom report that way.

If that last bit is close to what you want I can work up a bit more of a complete example if that would help but you can learn more by looking through the godocs for [SpecReport](https://pkg.go.dev/github.com/onsi/ginkgo/v2@v2.0.0/types#SpecReport) and [Report](https://pkg.go.dev/github.com/onsi/ginkgo/v2@v2.0.0/types#Report).

austince:
hey @onsi, thanks for the quick and thoughtful response.

[...]

PR #1009, opened 2022-07-22
Fix ReportAfterSuite usage in docs

austince:
While finally getting around to implementing https://github.com/onsi/ginkgo/issues/873 (which works perfectly, btw), found a few small
issues with the docs. Lmkwyt


onsi:
thanks for taking the time to fix the docs!

PR #1025, opened 2022-08-24
Synchronize AddReportEntries

### LINKED ISSUE #1020: Data race when adding reports from multiple goroutines
What I'm doing now, which may be completely misguided:

* For long-running tests, I want to periodically poll some system information and add it to the spec reports.
* I do this by spawning a goroutine in a `BeforeEach`, and closing it in the `AfterEach`
  ```go
  func ReportPeriodically(interval time.Duration) {
    mu := sync.Mutex{}
	cancelMap := map[string]func(){}

	ginkgo.BeforeEach(func() {
		report := ginkgo.CurrentSpecReport()
		name := report.FullText()

		ctx, cancel := context.WithCancel(context.Background())

		mu.Lock()
		defer mu.Unlock()
		cancelMap[name] = cancel
		ticker := time.NewTicker(interval)
		go func() {
			for {
                      // this can run both cases since there is no priority selection, but should not matter 
				select {
				case now := <-ticker.C:
					reportName := fmt.Sprintf("%s-periodic-%d", name, now.Unix())
					ginkgo.AddReportEntry(reportName, "some system info")
				case <-ctx.Done():
					ticker.Stop()
					return
				}
			}
		}()
	})

	ginkgo.AfterEach(func() {
		report := ginkgo.CurrentSpecReport()
		name := report.FullText()
		mu.Lock()
		defer mu.Unlock()
		if cancel, ok := cancelMap[name]; ok {
			cancel()
		}
		delete(cancelMap, name)
	})
  }
  ```
* I run the suite **sequentially**
* Data Race:
  ```
  WARNING: DATA RACE
  Read at 0x00c0005824d0 by goroutine 18:
  github.com/onsi/ginkgo/v2/internal.(*Suite).AddReportEntry()
      /home/runner/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.1.4/internal/suite.go:226 +0x386
  github.com/onsi/ginkgo/v2.AddReportEntry()
      /home/runner/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.1.4/reporting_dsl.go:69 +0x275
  ```

It seems the suite doesn't lock the entries array for reads/writes: https://github.com/onsi/ginkgo/blob/aae4480d35116455d7115e71120566e8be55b4a4/internal/suite.go#L226

I'm wondering if there's a better way to accomplish this, or if you'd accept a PR that protects access to the `ReportEntries` with a `sync.RWMutex`?

austince:
Allows report entries to be added concurrently.

Closes #1020

Signed-off-by: austin ce <austin.cawley@gmail.com>


types/types.go
```diff
@@ -164,6 +164,8 @@ type SpecReport struct {
 	CapturedStdOutErr string
 
 	// ReportEntries contains any reports added via `AddReportEntry`
+	// `AddReportEntry` supports adding entries concurrently, but direct read
+	// access to the slice is not thread-safe.
 	ReportEntries ReportEntries
```
austince, line 169: Would have liked to synchronize both read and write access at this level, but couldn't think of how to do it without introducing breaking changes.

onsi:
this is, in fact, more complicated than i'd hoped.  i'm out of time this morning but will try to look at it and solve it again soon.

PR #1040, opened 2022-09-19
chore(docs): fix code blocks without language settings

Abirdcfly:
Just reading docs and find like #1038 and #1039.
I guess the tiny pr is more suitable.


onsi:
thanks!

Issue #1029, opened 2022-08-30
DescribeTable with no entries should fail


Airblader:
In our project we just ran into the following human error situation:

```
var _ = Describe("Some API", func() {
	DescribeTable("for some context", func() {
		DescribeTable("does something", func(someArg bool) {
			// ...
		},
			Entry("", true),
			Entry("", false),
		)
	})
})
```

These tests will never execute. Why? Because the outer `DescribeTable` should be a `Describe` instead. As-is, this will silently pass and never execute anything.

Thus, the question becomes: should `DescribeTable` with no test cases perhaps fail?

eriknyk:
you have nested `DescribeTable ` and since the first `DescribeTable` have not any valid `Entry` sentence it doesn't nothing.
Maybe a Warning from ginkgo could help.

onsi:
hey @Airblader this makes sense.  The only minor drawback I see is that users who are in the middle of developing their specs and have a placeholder `DescribeTable` in place that they haven't fleshed out yet will get a failure until they do...

...but honestly that's a small price to pay given the irritation and costliness of this particular silently-passing-without-running-anything failure mode.

I'll add it to the backlog and will try to get to it soon.

Airblader:
Thanks, @onsi! I'd probably argue that the case you're describing is even more reason to fail in this case. 😄 

There's always `XDescribeTable` to temporarily disable the test, too. Perhaps the error message when failing could point to that so users have a way forward (I know from my team that the X variants weren't actually known to everyone).

onsi:
Yep that makes sense!

PR #1043, opened 2022-09-22
Exit if error: describeTable with no parameters

felipe88alves:
Fix: https://github.com/onsi/ginkgo/issues/1029
[Backlog Item #183159888](https://www.pivotaltracker.com/story/show/183159888)

Proposal is to add validation of the length of the arguments passed to the DescribeTable function prior to any validation of the arguments itself.

There were no tests on equivalent DescribeTable errors (i.e.: MultipleEntryBodyFunctionsForTable and InvalidEntryDescription). I don't know if there's any history there or not, but following convention, I also did not add tests for this new error. I'll gladly take comments on that decision.

No document updates were made since this simply reflects current expected behaviour.

### DIFF
--- a/table_dsl.go
+++ b/table_dsl.go
@@ -135,6 +135,10 @@ func generateTable(description string, args ...interface{}) {
 		return "Entry: " + strings.Join(out, ", ")
 	}
 
+	if len(args) == 1 {
+		exitIfErr(types.GinkgoErrors.MissingParametersForTableFunction(cl))
+	}
+
 	for i, arg := range args {
 		switch t := reflect.TypeOf(arg); {
 		case t == nil:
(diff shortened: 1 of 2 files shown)


onsi:
thanks @felipe-alves-eric - yes these edges aren't particularly well-tested - but your approach here makes sense.  a function with no inputs is invalid and this catches the issue identified in #1029 nicely.

felipe88alves:
Thanks for the feedback @onsi! 
I really enjoyed contributing, would it be ok for me to continue taking tasks in the backlog to help out?
And if so, do you recommend any task to take?

onsi:
hey @felipe88alves - sorry for missing this.  Thanks for all the great work over on the [MustPassRepeatedly](https://github.com/onsi/ginkgo/pull/1051) decorator!

I'd love more help and if this flow of PRs/reviews is working for you we can think through a few other items.

PR #1044, opened 2022-09-26
Make the outline command able to use the DSL import

bcrochet:
The outline command throws an error if the only the core DSL is imported. This allows any of the DSL to be imported and still produce a valid outline.

Signed-off-by: Brad P. Crochet <brad@redhat.com>


onsi:
Great catch, thanks!

PR #1046, opened 2022-10-02
fixes crashes on newer Ruby 3 installations by upgrading github-pages…

thediveo:
… gem dependency

- upgrades github-pages gem dependency to 227.
- regenerates gem lock file.

See also jekyll/jekyll#9115.


onsi:
thanks!!

Issue #1034, opened 2022-09-01
The Opposite of `FlakeAttempts`


astrieanna:
There's a set of tests that we want to rerun 10 times per run of the suite; kind of the opposite of FlakeAttempts, since the goal is they have to run _and_ pass 10 times to count as passing.

The goal is to ensure that a setting that causes the code-under-test's output to be ordered is actually working, rather than the random order happening to align with sorted order sometimes.

```
// Test this 10 times to eliminate randomness
DescribeTable("Sorted",
  func(x, y, z SomeType) {
    // Expect stuff to be sorted
  },
  basicTests,
  // TODO find a better way than copy pasting
  basicTests,
  basicTests,
  basicTests,
  basicTests,
  basicTests,
  basicTests,
  basicTests,
  basicTests,
  basicTests,
)
```

Is there a better way to write that in Ginkgo currently?

My dream is a new decorator that does the opposite of `FlakeAttempts`, so maybe `RequiredPasses`. This has only come up once in our test suites that I know of (and thus isn't really worth investing time in improving), so I figured I'd layout the request to see if comes up for anyone else.

onsi:
Hey @astrieanna - this totally makes sense and is something I've done as well (usually as a `for` loop either around an `It` or in the body of an `It`).  There _is_ a CLI flag `ginkgo -repeat=N` which will repeat the entire suite (and you could pair with filters to just repeat one test) but really this should be a decorator like`FlakeAttempts`.

I'll add it to the backlog.

PR #1051, opened 2022-10-11
Draft: introduction of RepeatAttempts decorator

felipe88alves:
Fix: https://github.com/onsi/ginkgo/issues/1034
[Backlog Item #183169714](https://www.pivotaltracker.com/story/show/183169714)

Proposal is to add a new decorator to evaluate if a Spec continuously passes. The idea is that this test will fail if ANY of its iterations fails.

Most of the design decisions and subsequent tests were inspired by the FlakeAttempts decorator.
Documentation was also added near the FlakeAttempt due to their similar nature.

A couple of questions were raised while thinking about this problem, feedback on them would be appreciated:

#### Naming
The naming of the new decorator is a point of contention. The name RepeatAttempts was used temporarily, but I would prefer the following proposals (respectively):

1. `RepeatSpecs`: More descriptive of what decorator's action. Execute a specific Spec repeatedly.
2. `RequiredPasses`: More descriptive of the decorator's intent. That all the test must pass a set amount of times.
3. `RepeatAttempts`: Ties the functionality to that of the FlakeAttempts.

#### Interworking with FlakeAttemps decorator
A decision was made to forbid the use of both decorator's simultaneously. A new error has been created to handle the situation where both decorators are defined for an Spec Group. *Needs approval*

#### Interworking with ginkgo CLI --flakeattempts flag 
What should be the expected behavior when --flakeattempts is defined in CLI and the RepeatAttemps decorator is used?
I was inclined to add a new error here to prevent that possibility, but was fearful that it could be too much of a limitation.
*This has not been handled, and is thus an Open issue that needs to be solved prior to merging this PR.*

#### Interworking with ginkgo CLI --repeat flag
A decision was made to treat them as independent actions. The --repeat flag repeats the execution of the suite as a whole, whereas the RepeatAttempts decorator repeats the execution of a specific Spec(s). An entry was put into the docs to explain this relation (or lack thereof). *Needs approval*

#### default_reporter.go behaviour
What is the motivation for only reporting the `Flaked` section of the report only populated when the test passes?
Ref: reporters/default_reporter.go ~ line 320
the question will help answer if the `Repeated` section of the same report should be populated when the test passes or fails. Current assumption is to do so wen it passes, but I don't understand why not report the `Flake` and `Repeated` sections when the test passes or fails, as they would be complimentary to the `Passed` and `Failed` fields anyway.

The following questions are related to testing. Once again, feedback on them would be appreciated:
#### FlakeAttempts and RepeatAttempts differentiator in Report test
I needed to find a way to distinguish the numAttempts caused by FlakAttempts or RepeatedAttempts in the S() convenience helper function of the `default_reporter_test.go` and added a bool type to do so. Not the most elegant solution, but let me know what you think.
Ref: reporters/default_reporter_test.go ~ line 103

#### Negative test cases
Not sure how to do the test case for the repeated decorator in the decorations_fixture_suite_test.go file, as it would need to be a negative test case. Perhaps the other tests suffice and this can be removed. If not, I kindly ask for some guidance on how to do the negative tests here. The test is currently commented for now awaiting some guidance.
Ref: integration/_fixtures/decorations_fixture/decorations_fixture_suite_test.go ~ line 34

#### Skipped tests
I found a couple of FlakeAttempt test which are currently skipped in its suite. Should that be corrected?
Ref: integration/_fixtures/flags_fixture/flags_test.go ~ line 88


onsi:
Thanks for investing so much time in this!  I'll try to take a look tomorrow and get back to you!

reporters/default_reporter_test.go
```diff
@@ -109,6 +109,12 @@ func S(options ...interface{}) types.SpecReport {
 			report.AdditionalFailures = append(report.AdditionalFailures, option.(types.AdditionalFailure))
 		case reflect.TypeOf(0):
 			report.NumAttempts = option.(int)
+		case reflect.TypeOf(true):
```
onsi, line 112: I think, instead, we can have a new `type MaxFlakeAttempts int` and `type RequiredPasses int`

reporters/default_reporter_test.go
```diff
@@ -1026,6 +1032,18 @@ var _ = Describe("DefaultReporter", func() {
 			DELIMITER,
 			"",
 		),
+		Entry("a failing test that was retried",
+			C(),
+			S(CTS("A"), "B", CLS(cl0), cl1, 2, false, types.SpecStateFailed),
```
onsi, line 1037: S(CTS("A"), "B", CLS(cl0), cl1, 2, MaxFlakeAttempts(2), types.SpecStateFailed),

onsi: ^ I think that would be a bit clearer

types/types.go
```diff
@@ -151,10 +151,17 @@ type SpecReport struct {
...
 	Failure Failure
 
-	// NumAttempts captures the number of times this Spec was run.  Flakey specs can be retried with
-	// ginkgo --flake-attempts=N
+	// NumAttempts captures the number of times this Spec was run.
+	// Flakey specs can be retried with ginkgo --flake-attempts=N or the use of the FlakeAttempts decorator.
+	// Repeated specs can be retried with the use of the RepeatAttempts decorator
 	NumAttempts int
 
+	// IsFlakeAttempts captures whether the spec has been retried with ginkgo --flake-attempts=N or the use of the FlakeAttempts decorator.
```
onsi, line 159: As mentioned in the issue I think:

```go
MaxFlakeAttempts int
RequiredPasses int
```

might be more informative and easier to work with.

types/types.go
```diff
@@ -363,11 +370,22 @@ func (reports SpecReports) CountWithState(states SpecState) int {
...
 	for i := range reports {
-		if reports[i].State.Is(SpecStatePassed) && reports[i].NumAttempts > 1 {
+		if reports[i].IsFlakeAttempts && reports[i].State.Is(SpecStatePassed) && reports[i].NumAttempts > 1 {
+			n += 1
+		}
+	}
+	return n
+}
+
+// CountOfRepeatedSpecs returns the number of SpecReports that passed after multiple attempts
```
onsi, line 384: I don't think we need this - there's no need to tell the user that a spec was repeated if they asked for it.  Whereas we _do_ want to tell the user "hey, this suite passed but only because we retried some flaky specs". (hence `CountOfFlakedSpecs()`

onsi (commented):
left some thoughts on the commit!

internal/group.go
```diff
@@ -299,16 +302,39 @@ func (g *group) run(specs Specs) {
...
+				multipleExecutionDecorator = reflect.TypeOf(FlakeAttempts(0))
 				maxAttempts = g.suite.config.FlakeAttempts
+				g.suite.currentSpecReport.IsFlakeAttempts = true
+			}
+
+			if g.suite.currentSpecReport.IsRepeatAttempts {
+				multipleExecutionDecorator = reflect.TypeOf(RepeatAttempts(0))
+				maxAttempts = max(1, spec.RepeatAttempts())
+			} else if g.suite.config.FlakeAttempts > 0 {
+				//What should be the behavior when --flakeattempts is defined in CLI and the RepeatAttemps decorator is used?
```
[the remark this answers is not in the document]
felipe88alves: Answered with suggestions in the comments

internal/group.go
```diff
@@ -318,9 +344,17 @@ func (g *group) run(specs Specs) {
...
 
-				if g.suite.currentSpecReport.State.Is(types.SpecStatePassed | types.SpecStateSkipped | types.SpecStateAborted | types.SpecStateInterrupted) {
-					break
+				switch t := multipleExecutionDecorator; {
+				case t == reflect.TypeOf(FlakeAttempts(0)):
+					if g.suite.currentSpecReport.State.Is(types.SpecStatePassed | types.SpecStateSkipped | types.SpecStateAborted | types.SpecStateInterrupted) {
+						break maxAttemptForLoop
+					}
+				case t == reflect.TypeOf(RepeatAttempts(0)):
+					if g.suite.currentSpecReport.State.Is(types.SpecStateFailed | types.SpecStateSkipped | types.SpecStateAborted | types.SpecStateInterrupted) {
```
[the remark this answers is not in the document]
felipe88alves: What about `types.SpecStatePanicked`?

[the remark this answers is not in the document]
onsi: yes.  in fact... I think you want `if !g.suite.currentSpecReport.State.Is(types.SpecStatePassed) {` - since the idea is it has to pass cleanly each time.

reporters/default_reporter_test.go
felipe88alves: Should NumAttempts be removed then?
If we add these new type, then one of them will always match NumAttempts.

types/types.go
felipe88alves: Answered in comment discussion:

> Based on this feedback, I'll change the `CountOfRepeatedSpecs` to only populate the `Repeated` report field when the test with the `MustPassRepeatedly` decorator fails. This was the user can understand whether a test failure is associated with a `MustPassRepeatedly` decorator or not.

felipe88alves:
> Of these I prefer `RequiredPasses(3)` but would actually suggest `MustPassRepeatedly(3)`

Nice one. All references to the funcitonality have been modified to reflect chosen name: `MustPassRepeatedly`

> I think if `RepeatAttempts `is provided it overrides any attempt to use `FlakeAttempts:`
> - In the case of a `FlakeAttempts `decorator an error is raised (as you've done)
> - In the case of --flake-attempts the cli configured flake attempts is silently ignored for specs that have `RepeatAttempts`.

Ok. I'll add this information in the docs for reference. Should I also add a comment in the code, stating that `MustPassRepeatedly` takes precedence over the --flake-attempts cli config (line 318 of group.go)?

> Because if it fails then it is treated as a failure and appears as such. The intent is to convey "These specs passed, but only because they were retried."

> - If the spec passed on all attempts, it just passes. No need to emit anything in the reporter output. If a user is running with -v they should see "Ginkgo: Attempt https://github.com/onsi/ginkgo/issues/1 Passed. Repeating...". If they look at a JSON report they should see the number of repeats. (You already have this in place but I have a few tweaks to suggest)
> - If the spec fails on any attempt the failure is reported. Since the SpecReport tracks the number of repeats we can tack on the attempt to the [FAILED] headers [here](https://github.com/onsi/ginkgo/blob/822a937ec32e0b3751c474c815c7dce97396d75b/reporters/default_reporter.go#L175-L184). For example Ginkgo could say • [TIMEDOUT] During repetition #1. This isn't super important, though, since the GinkgoWriter output will be emitted and it will include all the details.

Based on this feedback, I'll change the `CountOfRepeatedSpecs` to only populate the `Repeated` report field when the test with the `MustPassRepeatedly` decorator fails. This was the user can understand whether a test failure is associated with a `MustPassRepeatedly` decorator or not.

> I'll comment on this in the commit but I think rather than two bools it would be useful to record the configuration used when the spec ran in the SpecReport. So rather than have types.SpecReport.IsFlakeAttempts and types.SpecReport.IsRepeatAttempts be bools we could have types.SpecReport.MaxFlakeAttempts and `types.SpecReport.RequiredPasses" be set to the appropriate integers.

+1, will work on this

> I think something like:
> 
> countRepeat := 0
> 
> ```
> It("fails eventually", func() {
>   	countRepeat += 1
>   	if countRepeat >=3 {
>   		Fail(fmt.Sprintf("failed on attempt #%d", countRepeat)
>   	}
>   }, RepeatAttempts(3))
> ```
> and then in the integration spec we have:
> 
> `Eventually(session).Should(gbytes.Say("failed on attempt #3"))`
> If you want you can also assert that the Ginkgo: Attempt #1 Passed. Repeating... lines appeared:
> 
> ```
> Eventually(session).Should(gbytes.Say("Ginkgo: Attempt #1  Passed. Repeating..."))
> Eventually(session).Should(gbytes.Say("Ginkgo: Attempt #2  Passed. Repeating..."))
> Eventually(session).Should(gbytes.Say("failed on attempt #3"))
> ```

Think I finally understood these _fixture-based tests 😄 
In any case, according to what I was seeing, the FlakeAttempts test was being skipped in the decorations_fixture, due to the Focus decorator also being tested in the same suite.

As a solution, I broke out the Focus and Pending and Offset decorators and the FlakeAttempts and MustPassRepeatedly decorators into different Test Suites. I also changed the decoration_test.go integration test to reflect this change

internal/group.go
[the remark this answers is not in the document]
felipe88alves: In that case ` if g.suite.currentSpecReport.State.Is(types.SpecStateFailureStates) {` might be cleaner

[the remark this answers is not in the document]
onsi: sure that works too :)

reporters/default_reporter_test.go
onsi: Yes I think that's right.

Given the phrasing I think the defaults would be `MaxFlakeAttempts = 0` and `RequiredPasses = 1` right? If we want the defaults to be zero it could be `RequiredAdditionalPasses`

onsi:
> Nice one. All references to the funcitonality have been modified to reflect chosen name: `MustPassRepeatedly`

Great :)

> Ok. I'll add this information in the docs for reference. Should I also add a comment in the code, stating that `MustPassRepeatedly` takes precedence over the --flake-attempts cli config (line 318 of group.go)?

Yes, I think it would make sense there as well.

> Based on this feedback, I'll change the `CountOfRepeatedSpecs` to only populate the `Repeated` report field when the test with the `MustPassRepeatedly` decorator fails. This was the user can understand whether a test failure is associated with a `MustPassRepeatedly` decorator or not. Is that ok or would you really prefer to not have any such feedback?

Got it - yeah I think what you're saying makes sense.

> Think I finally understood these _fixture-based tests 😄 In any case, according to what I was seeing, the FlakeAttempts test was being skipped in the decorations_fixture, due to the Focus decorator also being tested in the same suite.
> 
> As a solution, I broke out the Focus and Pending and Offset decorators and the FlakeAttempts and MustPassRepeatedly decorators into different Test Suites. I also changed the decoration_test.go integration test to reflect this change

OK thanks - I'll take a look!

felipe88alves:
> If the spec fails on any attempt the failure is reported. Since the SpecReport tracks the number of repeats we can tack on the attempt to the [FAILED] headers [here](https://github.com/onsi/ginkgo/blob/822a937ec32e0b3751c474c815c7dce97396d75b/reporters/default_reporter.go#L175-L184). For example Ginkgo could say • [TIMEDOUT] During repetition #1. This isn't super important, though, since the GinkgoWriter output will be emitted and it will include all the details.

I've noticed that the report for the Successful FlakedTests contains this Retry Denoter. I don't know what this retry denoter is meant to represent, but do we also need it for the report of the failed Repeated tests?

onsi:
it's just representing that the spec was flaky and had to be rerun.  I don't think we need it for retry.  let me know if you think things are in a good place and I can take another look tonight and see if we can get this merged in. so you know:  I'm out of town the next few days and won't be at my computer much - hopefully we can get it in before then, if not we'll aim for early next week.

felipe88alves:
Just reviewing the last changes. Hoping to push it up tonight.

docs/index.md
```diff
@@ -4839,13 +4839,13 @@ With this setup, `"is flaky"` and `"is also flaky"` will run up to 3 times.  `"i
 
 If `ginkgo --flake-attempts=N` is set the value passed in by the CLI will override all the decorated values.  Every test will now run up to `N` times.
 
-#### The RepeatAttempts Decorator
-The `RepeatAttempts(uint)` decorator applies to container and subject nodes.  It is an error to apply `RepeatAttempts` to a setup node.
+#### The MustPassRepeatedly Decorator
+The `MustPassRepeatedly(uint)` decorator applies to container and subject nodes.  It is an error to apply `MustPassRepeatedly` to a setup node.
 
-`RepeatAttempts` allows the user to flag specific tests or groups of tests for debbuging flaky tests.  Ginkgo will run tests up to the number of times specified in `RepeatAttempts` until they fail.  For example:
+`MustPassRepeatedly` allows the user to flag specific tests or groups of tests for debbuging flaky tests.  Ginkgo will run tests up to the number of times specified in `MustPassRepeatedly` until they fail.  For example:
```
onsi, line 4845: This is a bit of a nitpick (sorry!) but I've been trying to call "tests" in Ginkgo "specs"  throughout the documentation (though I'm sure I've missed quite a few!). Would you be up for changing the language here from test/tests to spec/specs?

The rationale is to distinguish Ginkgo specs from regular go tests - but it's mostly about consistency tbh.

internal/group.go
```diff
@@ -332,8 +330,8 @@ func (g *group) run(specs Specs) {
 					switch t := multipleExecutionDecorator; {
```
onsi, line 330: rather than tracking `multipleExecutionDecorator` and doing a type switch could these be `if g.suite.currentSpecReport.MaxFlakeAttampts > 0`  and `else if g.suite.currentSpecReport.MaxMustPassRepeatedly > 0`?

reporters/default_reporter.go
```diff
@@ -149,10 +149,8 @@ func (r *DefaultReporter) DidRun(report types.SpecReport) {
 			}
 		} else {
 			header, stream = denoter, true
-			if report.NumAttempts > 1 {
-				if report.IsFlakeAttempts {
-					header, stream = fmt.Sprintf("%s [FLAKEY TEST - TOOK %d ATTEMPTS TO PASS]", r.retryDenoter, report.NumAttempts), false
-				}
+			if report.MaxFlakeAttempts > 1 {
```
onsi, line 152: I don't think this is right.  You need `report.NumAttempts >  1 && report.MaxFlakeAttempts > 1` otherwise every spec marked with `MaxFlakeAttempts(N)` will get this output.

The intent is to only tell the user "hey this spec was flakey" if it actually takes more than one try to succeed.  Otherwise they should just see a regular passing spec.

reporters/default_reporter.go
```diff
@@ -176,11 +174,7 @@ func (r *DefaultReporter) DidRun(report types.SpecReport) {
 		}
 	case types.SpecStateFailed:
 		highlightColor, header = "{{red}}", fmt.Sprintf("%s [FAILED]", denoter)
```
onsi, line 176: I think this line is leftover form an earlier merge?  You should be able to remove it as highlightColor is set elsewhere now

reporters/default_reporter.go
```diff
@@ -191,6 +185,9 @@ func (r *DefaultReporter) DidRun(report types.SpecReport) {
 		header = fmt.Sprintf("%s! [ABORTED]", denoter)
 	}
 
+	if report.State.Is(types.SpecStateFailureStates) && report.MaxMustPassRepeatedly > 1 {
```
onsi, line 188: 👍

reporters/default_reporter_test.go
```diff
@@ -107,14 +109,12 @@ func S(options ...interface{}) types.SpecReport {
...
-		case reflect.TypeOf(0):
+		case reflect.TypeOf(report.NumAttempts):
 			report.NumAttempts = option.(int)
-		case reflect.TypeOf(true):
-			if option.(bool) {
-				report.IsFlakeAttempts = true
-			} else {
-				report.IsRepeatAttempts = true
-			}
+		case reflect.TypeOf(FlakeAttempts(0)):
```
onsi, line 114: nice - I think this is much clearer.  thanks!

onsi (commented):
Looks good - I left a couple of comments but I think once they're resolved we'll be able to merge this in.  I'm going to be out of town for the next few days but can pull the PR in and cut a new release shortly after I get back!

felipe88alves:
At this point all review items have been addressed apart from: https://github.com/onsi/ginkgo/pull/1051#discussion_r995307891
Pending conclusion on design.

onsi:
he @felipe88alves - I'm back from vacation.  This all looks pretty good.  It looks like a merge conflict has popped up so if you a have a minute to take a look that would be great.

On [this](https://github.com/onsi/ginkgo/pull/1051#discussion_r995307891) I think I'd prefer to avoid switching on the type check since there is already context available in the code via the parameters stored on the `currentSpecReport` to decide the switch.  I'd be happy to pull this in and just make that tweak myself, though if you'd prefer.

PR #1067, opened 2022-10-21
Add GinkgoLogr

### LINKED ISSUE #1063: Implement logr.LogSink for GingkoWriter
Hey :wave:,

I use [logr](https://github.com/go-logr/logr) in some projects and use Ginkgo for testing. GingkoWriter offers the option to record some output of failing test. Implementing `LogSink` allows failing Ginkgo tests to outright record what is to be logged anyway, which helps investigating the cause. Adding the glue code would be convenient.

Instead of implementing `LogSink` directly, the implementation could be wrapped in an other struct. As I have a basic implementation around, I may work on this.

Nuckal777:
Resolves #1063 .

### DIFF
--- a/core_dsl.go
+++ b/core_dsl.go
@@ -23,6 +23,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/go-logr/logr"
 	"github.com/onsi/ginkgo/v2/formatter"
 	"github.com/onsi/ginkgo/v2/internal"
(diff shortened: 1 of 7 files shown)


onsi:
hey there this looks good.  can you also update the docs (probably here is best: https://github.com/onsi/ginkgo/blob/master/docs/index.md#logging-output)

to mention `GinkgoLogr`?

Thanks!

onsi:
Looks good - thanks!

PR #1095, opened 2022-12-14
Override environment formatter colors from envvars

blacksails:
This allows a user to set environment variables
`GINKGO_CLI_COLOR_<color>` with asci code overrides. This helps a lot when the colorscheme that you are using doesn't play well with the colors from the 256 color scheme which currently is hard coded.

Example `GINKGO_CLI_COLOR_RED=$'\x1b[31m' ginkgo` will use the red from the 8bit colors configured in the terminal.

### DIFF
--- a/formatter/formatter.go
+++ b/formatter/formatter.go
@@ -50,6 +50,15 @@ func NewWithNoColorBool(noColor bool) Formatter {
 }
 
 func New(colorMode ColorMode) Formatter {
+	getColor := func(color, defaultEscapeCode string) string {
+		color = strings.ToUpper(strings.ReplaceAll(color, "-", "_"))
+		envVar := fmt.Sprintf("GINKGO_CLI_COLOR_%s", color)
+		if escapeCode := os.Getenv(envVar); escapeCode != "" {
+			return escapeCode
+		}
+		return defaultEscapeCode
+	}
+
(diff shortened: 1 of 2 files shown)


blacksails:
I dont nessesarily expect this to get merged as-is, but I would love an option to be able to change the colors of the terminal output.

The formatter testsuite currently fails if environment variables are overridden. I guess the before each needs to unset the environment variables to ensure that the tests give the expected results.

onsi:
hey there - longer term my plan is to add support for some sort of .ginkgoconfig file or something that would be a natural place for something like this.  For now environment variables seem fine but I'd like to reserve the right to deprecate them at some point and transition to a config  file based approached.

Happy to pull this in once CI goes green, with those caveats.

blacksails:
@onsi Yea this could also be better implemented.

The following values would be nice to map in order to make configuration way easier.

`red` => `\x1b[31m`
`1` => `\x1b[31m`
`bright-red` => `\x1b[31;1m`
`8` => `\x1b[31;1m`
`16` => `\x1b[38;5;16m`

onsi:
hey, i think that's fine - but perhaps i'm not following?  are you planning on another PR to support those aliases?

blacksails:
@onsi yea will do it today.

PR #1130, opened 2023-02-01
test: update matrix for Go 1.20

blgm:


onsi:
hmm looks like that might be a real issue - i can poke at it later this week

blgm:
@onsi I've pushed a change to the test to "fix" the issue. It was relying on getting a very specific coverage number (71.4%) and that's now changed in Go 1.20 (to 80%). I guess increased coverage is good, and I'm highly sceptical of relying on coverage numbers other than 100% as an optimiser can alter the number of statements and change the result.

The changed code will accept either 71.4% or 80%. I'm a little uncomfortable with such an unsophisticated fix, but we are testing integration with Go coverage rather than how the numbers are calculated. I'll hold off merging in case you have thoughts.

onsi:
hey @blgm all your caveats make sense and that fix sounds good to me.  thanks for fixing it!

PR #1147, opened 2023-02-23
Add OmitSuiteSetupNodes to JunitReportConfig

### LINKED ISSUE #1145: Add new option to JunitReportConfig
My suggestion is to add to the JunitReportConfig a new option: "OmitSetupNodes"
that when is set to true, the Junit report won't create <testcase>s for setup nodes such as BeforSuite and AfterSuite.
for example: 
`<testcase name="[BeforeSuite]" classname="CNF Features e2e integration tests" status="passed" time="0"></testcase>`

This feature could be helpful for me because in [this](https://github.com/openshift-kni/cnf-features-deploy/tree/master) repo, we have a script that takes the output of the JUnit XML, and convert it to an MD file with all the tests cases, where we want to omit the Before/After suites. 

Currently, I see there is the "OmitLeafNodeType" option but it just leaves me with an empty test case like this one:
`<testcase name="" classname="CNF Features e2e integration tests" status="passed" time="0"></testcase>`

liornoy:
This PR adds a new option to the JunitReportConfig that when enabled prevents the creation of
testcase entries for setup nodes in the JUnit report.

fix #1145

### DIFF
--- a/reporters/junit_report.go
+++ b/reporters/junit_report.go
@@ -36,6 +36,9 @@ type JunitReportConfig struct {
 
 	// Enable OmitLeafNodeType to prevent the spec leaf node type from appearing in the spec name
 	OmitLeafNodeType bool
+
+	// Enable OmitSuiteSetupNodes to prevent the creation of testcase entries for setup nodes
+	OmitSuiteSetupNodes bool
 }
 
 type JUnitTestSuites struct {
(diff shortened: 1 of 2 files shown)


onsi:
thanks! this LGTM and i'll merge it in once CI finishes

onsi:
merged!  i should be cutting a release that includes this sometime next week.

PR #1192, opened 2023-05-03
fix hang with ginkgo -p

### LINKED ISSUE #1191: ginkgo run in parralell can hang when child command outlives test suite
I have been debugging hangs in ginkgo v2 in over last weeks and I think I finally found the problem.
I created https://github.com/Luap99/ginkgo-hang for a simple reproducer.

This diff seems to fix the issue for me:
```diff
diff --git a/internal/output_interceptor_unix.go b/internal/output_interceptor_unix.go
index f5ae15b..70d5647 100644
--- a/internal/output_interceptor_unix.go
+++ b/internal/output_interceptor_unix.go
@@ -26,6 +26,11 @@ func (impl *dupSyscallOutputInterceptorImpl) CreateStdoutStderrClones() (*os.Fil
        stdoutCloneFD, _ := unix.Dup(1)
        stderrCloneFD, _ := unix.Dup(2)
 
+       flags, _ := unix.FcntlInt(uintptr(stdoutCloneFD), unix.F_GETFD, 0)
+       unix.FcntlInt(uintptr(stdoutCloneFD), unix.F_SETFD, flags|unix.FD_CLOEXEC)
+       flags, _ = unix.FcntlInt(uintptr(stderrCloneFD), unix.F_GETFD, 0)
+       unix.FcntlInt(uintptr(stderrCloneFD), unix.F_SETFD, flags|unix.FD_CLOEXEC)
+
        // And then wrap the clone file descriptors in files.
        // One benefit of this (that we don't use yet) is that we can actually write
        // to these files to emit output to the console even though we're intercepting output
```

By using CLOEXEC we make sure the fds are never leaked into commands that are executed in the test suite.
I am happy to open a PR if you agree that this is the right approach to fix the problem.

Luap99:
When you run any child process that stay around longer than the test
suite ginkgo currently hangs. This is because the dup stdout/err fds are
leaked into the child thus read() will block on it as there is at least
one process still having the write pipe open.

From ginkgos POV it looks like it is done, you see the ginkgo result
output printed but then it just hangs and doe snot exit because of it.

To fix it we set FD_CLOEXEC on the dup-ed fds, this prevents them from
ever being leaking into other processes that could outlive the suite.
There is a dup3() call the could be uses to set the CLOEXEC option
directly but this seem linux only and thus is less portable.
The fcntl call should be good enough here, we do not need to be worried
about the race conditions described in the man page.
Ideally we should do some error handling in that function for both the
fcntl calls and the existing dup() above, however this seems like a
rather big change.

I am not so sure about how to test it properly, I added a test which
just executes `ginkgo run -p` and a test which only starts `sleep 60`.
Ginkgo then should exit right way and keep this process running. Then I
just make sure the gingo exits in under 15 seconds. As long as it is
below 60s it should fulfil the purpose.

Fixes #1191

Also see the other two commits for some test fixes I notcied

### DIFF
--- a/TODO
+++ b/TODO
@@ -1,0 +1,60 @@
+# Backlog
+[] @G fix hang with ginkgo -p 
+    https://github.com/onsi/ginkgo/issues/1192
+[] @B biloba needs to support "McDonald's"
+[] @G Document order of execution for setup nodes on the same level
+    https://github.com/onsi/ginkgo/issues/1172
+[] @G fail fast may cause Serial spec or cleanup Node interrupted
+    https://github.com/onsi/ginkgo/issues/1178
+
+# Needs-Prioritization
+[] @G Pull JUnit config out of code and into flags/config files
+    https://github.com/onsi/ginkgo/issues/1115
+[] @G -exec support for `ginkgo run`
+    https://github.com/onsi/ginkgo/issues/869
+[] @G Add `indent` to gcustom
+[] @G HaveField should not panic in FailureMessage if the field is non-existant.  It should return a meaningful failure message.
+[] @G allow NodeTimeout and GracePeriod to live on containers and propagate down
+[] @G Clean up ReportEntry decoding:
+    func (entry ReportEntry) DecodeValue(p interface{}) error {
+        // if raw is assignable to p, assign
+        // else - parse AsJSON
+    }
+[] @B Biloba lets you get all innerHTML and emit it on failure
+[] @B equivalent of puppeteer text selector?
+[] @B how do we invoke async functions?  what does await look like for those?  maybe time to actually read?  goal: remove the separate muhasibPostLogin function.
+[] @B https://github.com/onsi/biloba/issues/2
+[] @B right now polling an element fails if the browser if ollowing redirects.  so i'm using Eventually(b.Location) - instead of just Eventually("#el").Should(b.Exist()).  I think we need a more robust way to ensure biloba.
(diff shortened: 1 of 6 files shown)


onsi:
thanks for pulling this together - it looks like CI failed but i'll merge this in later this morning and take a look at the testing strategy.  i think your approach is the right one (spawn a sleep and ensure ginkgo exits in a timely manner) - but i need to think about where the test should live (right now it's in an existing fixture - whereas i think it might be better as a distinct fixture with the test itself living in ginkgo's integration suite).

in any event i'll take a look in a couple hours and cut a release soon

Luap99:
Yeah I wasn't sure about the test structure. Also seems like CI doesn't have access to the ginkgo binary as it seem to be run with go run not sure how to address that.

onsi:
no worries - that piece will be resolved when i move the test into the integration tests (they compile their own version of ginkgo to ensure we're testing against the actual code in the repo).  testing a testing framework can be... interesting ;)

onsi:
looks good now.  gonna pull this in.  thanks so much!

TODO
```diff
@@ -0,0 +1,60 @@
+# Backlog
```
Luap99, line 1: thanks for merging, did you intent to commit the TODO file?

onsi: no, sigh - i had just renamed it and updated my .gitignore on main but accidentally committed it in the pr.  i've cleaned it up now, though it will live on forever in the git history.  there's nothing sensitive in there

PR #1208, opened 2023-05-26
Fix return value in index.md

arosh:
Please forgive me if my understanding is incorrect.


onsi:
good catch, thanks for updating these.

Issue #1265, opened 2023-08-23
Configure MustPassRepeatedly at suite level


maxcleme:
We are looking at a way of configuring `MustPassRepeatedly` programatically at suite level before starting it, however it is not possible at the moment. 

Ideally, something similar to `FlakeAttempts`, where it can be set at node level using the decorator but also at [suite level ](https://pkg.go.dev/github.com/onsi/ginkgo/v2/types@v2.10.0#SuiteConfig.FlakeAttempts).

Any insight on how we could perform this? Is there any reason why it is not present at suite level?

onsi:
hey - you can use `ginkgo -repeat=N`  and the CLI will orchestrate repeats and require them all to pass

maxcleme:
@onsi Thanks for fast answer. 

Sadly we aren't using the CLI, I forgot to mention that! 

We'd like to set `MustPassRepeatedly` programatically before calling `RunSpecs`.

onsi:
That isn’t supported yet (it’s just not implemented).  i can imagine simply passing the decorator into `RunSpecs` would decorate all the nodes with `MustPassRepeatedly` and it would just work.  If you want to work on a PR that does that that would be great.

Can you share why you aren't using the CLI?  It’s required for running specs in parallel and I’d love to learn what issues you’ve bumped into.

maxcleme:
Works fine so I'll close! 

On a side note, any idea how we could get the timing of all these N executions for each tests during reporting step (`ReportAfterSuite("", func(report Report) {})`) ? I guess I could rebuild the timeline with `SpecEvents` but it looks like reinventing the wheel 🤔

onsi:
hey @maxcleme - the individual times aren't stored by Ginkgo but you can use [Report Entries](https://onsi.github.io/ginkgo/#attaching-data-to-reports) do something like this at the top level (e.g. right after the `TextX` function that invokes `RunSpecs()`:

```go
var _ = BeforeEach(func() {
    t := time.Now()
    DeferCleanup(func() {
		AddReportEntry(fmt.Sprintf("Run %d", CurrentSpecReport().NumAttempts, time.Since(t))
    })
})
```

If you share more about how you post-process the specs to generate reports I can offer some examples for how to then use these report entries.  But, it's basically just data and data transformation at that point.

maxcleme:
[...]

PR #1266, opened 2023-08-23
feat: allow MustPassRepeatedly decorator to be set at suite level

maxcleme:
Relates to https://github.com/onsi/ginkgo/issues/1265

Aims to allow `MustPassRepeatedly` to be set at suite level programatically, similarly of `FlakeAttempts`.

### DIFF
--- a/integration/repeat_test.go
+++ b/integration/repeat_test.go
@@ -98,4 +98,15 @@ var _ = Describe("Repeat", func() {
 			Ω(session.Err).Should(gbytes.Say("--repeat and --until-it-fails are both set"))
 		})
 	})
+
+	Context("if MustPassRepeatedly is set at suite config level", func() {
+		BeforeEach(func() {
+			fm.MountFixture("config_override_must_pass_repeatedly")
+		})
+
+		It("it should override node decorator", func() {
+			session := startGinkgo(fm.PathTo("config_override_must_pass_repeatedly"))
+			Eventually(session).Should(gexec.Exit(0))
+		})
+	})
 })
--- a/integration/_fixtures/config_override_label_filter_fixture/config_override_fixture_suite_test.go
+++ b/integration/_fixtures/config_override_label_filter_fixture/config_override_fixture_suite_test.go
@@ -1,4 +1,4 @@
-package config_override_fixture_test
+package config_override_label_filter_fixture_test
 
 import (
 	"testing"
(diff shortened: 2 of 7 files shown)


maxcleme:
Regarding the test, I'm not which way I should go

+ either add a new cases in `integration/_fixtures/decorations_fixture/flaky_repeated/flaky_repeated_fixture_suite_test.go`, using suite config instead of node decorator
+ edit `integration/_fixtures/config_override_fixture/config_override_fixture_suite_test.go` (which seems relevant?)
+ same as `internal/internal_integration/config_flake_attempts_test.go` ?

@onsi any thoughts? 😁

onsi:
i’ll try to take a closer look today but to unblock you:
- i’d definitely like an internal integration test - and it would be good to ensure that the override at the suite level plays nicely with any per-test values (happy to think through what the behavior should be with you.  at first blush i’d say it should do whatever FlakeAttempts does.)
- the outside-in integration test is less crucial.  one of the two you point out should be enough

integration/repeat_test.go
```diff
@@ -98,4 +98,15 @@ var _ = Describe("Repeat", func() {
 			Ω(session.Err).Should(gbytes.Say("--repeat and --until-it-fails are both set"))
 		})
 	})
+
+	Context("suite config overrides MustPassRepeatedly decorator", func() {
```
maxcleme, line 102: I wasn't really sure where to put that one, between the following places:
+ `integration/repeat_test.go`
+ `integration/decorations_test.go`
+ `integration/flags_test.go` (where lies the test for label filter suite override)

wdyt?

maxcleme:
@onsi I've added some tests, let me know what you think!

integration/repeat_test.go
```diff
@@ -98,4 +98,15 @@ var _ = Describe("Repeat", func() {
 			Ω(session.Err).Should(gbytes.Say("--repeat and --until-it-fails are both set"))
 		})
 	})
+
+	Context("if MustPassRepeatedly is set at suite config level", func() {
```
maxcleme, line 102: I wasn't really sure where to put that one, between the following places:
+ `integration/repeat_test.go`
+ `integration/decorations_test.go`
+ `integration/flags_test.go` (where lies the test for label filter suite override)

wdyt?

onsi: hey - i think this is fine here, thanks.

onsi:
i just cut a release with this in it, thanks!

PR #1275, opened 2023-09-19
Print logr prefix if it exists

Nuckal777:
Fixing an other small oversight in the logr integration.


onsi:
LGTM -  thanks!  I’ll cut a release later today.

PR #1283, opened 2023-10-09
`gofmt` examples in README

jtarchie:
As an open-source maintainer, I know how helpful having valid working examples in the documentation can be. This is an attempt to update the examples to be copy-pasted.

This transformation was part programmatic, part manual editing. I wrote a little Ruby script that took the code example between the triple-backticks, then tried various "fixes".

- add a package name
- replace placeholder `...` and `???` with multiline comment fenceposts
- try `go fmt`

If it could be formatted, I tried to resolve the issue manually until it could, or I decided I could not resolve it for reasons.

This was a quick pass. Not meant to be definitive.

```ruby
# frozen_string_literal: true

require 'tempfile'

filename = ARGV[0]
contents = File.read(filename)

contents.gsub!(/(?<=```go\n)(.*?)(?=```\n)/mu) do |original_code_block|
  code_block = original_code_block
  

  unless code_block.match(/^package \w+$/)
    # puts '  🚧 adding default `package books_test`'
    code_block = "package books_test\n#{code_block}"
  end

  filler = /\s+(\.\.\.|\?\?\?)\s+/
  if code_block.match(filler)
    # puts '  🚧 cleaning up filler'
    code_block = code_block.gsub(filler, '/* \1 */')
  end

  file = Tempfile.new(%w[main go])
  file.write(code_block)
  file.close

  skipper = /INVALID|is equivalent to/
  if original_code_block.count("\n") <= 3 || original_code_block.match(skipper)
    # puts '  ✅ valid go'
    original_code_block
  elsif system("gofmt -w #{file.path}")
    # puts '  ✅ valid go'
    File.read(file.path)
  else
    puts code_block
    puts '  🔴 could not format file invalid go'
    original_code_block
  end
end

File.write('errors.md', contents)
```

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -95,14 +95,14 @@ This will generate a file named `books_suite_test.go` in the `books` directory c
 package books_test
 
 import (
-  . "github.com/onsi/ginkgo/v2"
-  . "github.com/onsi/gomega"
-  "testing"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"testing"
 )
 
 func TestBooks(t *testing.T) {
-  RegisterFailHandler(Fail)
-  RunSpecs(t, "Books Suite")
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Books Suite")
 }
 ```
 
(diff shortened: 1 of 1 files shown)


jtarchie:
Thanks for the feedback, not ignoring, just on vacation. 😄 

I will explore your proposal here. Any leads on where the CSS I need to edit would be?

onsi:
No worries, obviously!  Hope vacation's going/gone well :)

I think it probably goes here somewhere:

https://github.com/onsi/ginkgo/blob/74bbd6552ca29d47cf6c828aac38d42fbd912bd9/docs/css/layout.css#L62-L80

PR #1302, opened 2023-11-11
Avoid allocations with `(*regexp.Regexp).MatchString`

Juneezee:
We should use `(*regexp.Regexp).MatchString` instead of `(*regexp.Regexp).Match([]byte(...))` when matching string to avoid unnecessary `[]byte` conversions and reduce allocations. A one-line change for free performance improvement.

Example benchmark:

```go
var goTestRegExp = regexp.MustCompile(`_test\.go$`)

func BenchmarkMatch(b *testing.B) {
	for i := 0; i < b.N; i++ {
		if match := goTestRegExp.Match([]byte("file_test.go")); !match {
			b.Fail()
		}
	}
}

func BenchmarkMatchString(b *testing.B) {
	for i := 0; i < b.N; i++ {
		if match := goTestRegExp.MatchString("file_test.go"); !match {
			b.Fail()
		}
	}
}
```

Result:

```
goos: linux
goarch: amd64
pkg: github.com/onsi/ginkgo/v2/ginkgo/watch
cpu: AMD Ryzen 7 PRO 4750U with Radeon Graphics
BenchmarkMatch-16          	 5665784	       314.4 ns/op	      16 B/op	       1 allocs/op
BenchmarkMatchString-16    	 8481872	       140.5 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/onsi/ginkgo/v2/ginkgo/watch	4.321s
```

### DIFF
--- a/ginkgo/internal/test_suite.go
+++ b/ginkgo/internal/test_suite.go
@@ -226,7 +226,7 @@ func suitesInDir(dir string, recurse bool) TestSuites {
 	files, _ := os.ReadDir(dir)
 	re := regexp.MustCompile(`^[^._].*_test\.go$`)
 	for _, file := range files {
-		if !file.IsDir() && re.Match([]byte(file.Name())) {
+		if !file.IsDir() && re.MatchString(file.Name()) {
 			suite := TestSuite{
 				Path:        relPath(dir),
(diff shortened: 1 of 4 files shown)


onsi:
SGTM thanks!

PR #1333, opened 2024-01-08
Add GinkgoTB() function

ebabani:
This function returns a wrapper around GinkgoT() which satisfies the `testing.TB` interface.

- Update fail tests with tests for GinkgoT() and GinkgoTB().
- Fix a bug where GinkgoT().Fail() was not reporting the correct line on error.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -5478,6 +5478,8 @@ When using Gomock you may want to run `ginkgo` with the `-trace` flag to print o
 
 `GinkgoT()` also provides additional methods that are Ginkgo-specific.  This allows rich third-party integrations to be built on top of Ginkgo - with GinkgoT() serving as a single connection point.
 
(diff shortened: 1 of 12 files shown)


onsi:
thanks!  this looks great - but I think there’s a bug in the (previous, existing) test that checks that all definitions in the core DSL are exported in the modular DSL package (long story….).  I’ll take a look later today/tomorrow and fix it.

ebabani:
> thanks! this looks great - but I think there’s a bug in the (previous, existing) test that checks that all definitions in the core DSL are exported in the modular DSL package (long story….). I’ll take a look later today/tomorrow and fix it.

I pushed a fix for that failure. Forgot to run those tests and was only running integration tests.

I updated the symbol check to ignore functions with a receiver. Those methods are never going to cause an import conflict, and I added the missing package level functions and types to `core_dsl.go`.

onsi:
fantastic, thank you! i’ll cut a release soon :)

PR #1337, opened 2024-01-16
bugfix(docs): use Unsetenv instead of Clearenv

sxllwx:
The documentation incorrectly uses `os.Clearenv` to clear specific environment variables. In fact, we should use `os.Unsetenv`.

The APIs related to these two functions are as follows:

https://pkg.go.dev/os#Clearenv
https://pkg.go.dev/os#Unsetenv


onsi:
thanks!

Issue #1114, opened 2023-01-05
GinkgoRecover not aborting the current It test node when Expect in goroutine fails


thediveo:
`go.mod`:
```
github.com/onsi/ginkgo/v2 v2.6.1
github.com/onsi/gomega v1.24.2
```

```go
It("renders user tree with loose threads with CLI -d", func() {
		By("creating a stray task with its own namespace...")
		tidch := make(chan int)
		done := make(chan struct{})
		go func() {
			defer GinkgoRecover()
			// unix.Unshare will fail with err != nil
			Expect(unix.Unshare(unix.CLONE_FS | unix.CLONE_NEWNS)).To(Succeed())

			tidch <- unix.Gettid()
			<-done
		}()

		tid := <-tidch
		// ...
}
```

When running this test (as a non-root user) the expectation in the separate go routine will fail. However, the test node itself won't be aborted but instead will "hang" until the overall test watchdog barks and bites.

This is probably the, erm, _expected_ behavior (pun intended). What is the proper way to correctly handle failed expectations while the original test node go routine is blocked on some channel read? Somehow I would imagine using a test context and then cancel this? Is there a convenient way to ensure that the context is only cancelled when GinkgoRecover actually meets with a failed expectation (panic)?

onsi:
I think in this case I would recommend:

```go
It("renders user tree with loose threads with CLI -d", func() {
		By("creating a stray task with its own namespace...")
		tidch := make(chan int)
		done := make(chan struct{})
		go func() {
			defer GinkgoRecover()
			// unix.Unshare will fail with err != nil
			Expect(unix.Unshare(unix.CLONE_FS | unix.CLONE_NEWNS)).To(Succeed())

			tidch <- unix.Gettid()
			<-done
		}()

		var tid int
                Eventually(tidch).Should(Receive(&tid))
		// ...
}
```

that's probably the cleanest/simplest/most legible way.  No doubt there are other potential approaches (e.g. additional `defer`s in the goroutine or using a `SpecTimeout` to kill the whole test if it hangs) but I think this approach will give you the cleanest failure message.

thediveo:
Ah, that's a nifty solution!

Alternatively, I was thinking about either allowing `GinkgoRecover()` to accept an optional `conext.CancelFunc`? Or instead offering a new `GinkgoRecoverAndCancel()`. However, this needs instead more test harness code in order to create  a cancel-able context and bailing out while waiting for the receive (that never comes).

[...]

PR #1339, opened 2024-01-18
Fix docs for handling failures in goroutines

austince:
Finally got around to https://github.com/onsi/ginkgo/issues/1114#issuecomment-1879291376 !

- Fix goroutine failure handling docs
- Correct CSS rules for desktop


onsi:
perfect, thanks!

Issue #1329, opened 2024-01-05
Merging coverage profiles leads lots of duplicated lines and uses lots of memory


myzhan:
I'm using ginkgo to run the testcases in a project that has 26w lines of code, with the -p flag enabled.

Issues
1. When running testcases in parallel, more cpu cores lead to more duplicated lines. If I add the -p flag and run on a 64 cores machine, the size of cover.out file is 7.9GB. And the size of cover.out file is 130MB when -p flag removed.
2. All the coverage files are read into combined buffer, which use lots of memory. And ginkgo will be oom-killed.

myzhan:
OK, I'll find some time to submit a PR.

PR #1340, opened 2024-01-20
merge coverages instead of combining them (#1329)

myzhan:
I copy [gocovmerge](https://github.com/shabbyrobe/gocovmerge) to ginkgo, because it hasn't versions, and the code is easy to maintain.

### DIFF
--- a/ginkgo/internal/gocovmerge.go
+++ b/ginkgo/internal/gocovmerge.go
@@ -1,5 +1,29 @@
+// Copyright (c) 2015, Wade Simmons
+// All rights reserved.
+
+// Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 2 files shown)


onsi:
Sorry for the delay - thanks this looks good to me.  I only have one request, can you include the original license:
https://github.com/wadey/gocovmerge/blob/master/LICENSE

by adding it to the comment at the top of the new `gocovmerge.go` file?  It’s a BSD-2 simplified license so I believe the two are largely compatible but I think it’s best to list it explicitly and make sure Wade Simmons gets the copyright notice. 

@wadey please let us know if you have any concerns!

onsi:
perfect, thanks!

PR #1352, opened 2024-02-07
chore: test with Go 1.22

blgm:
- Rather than hard-coding the Go version in the workflow, we will use the `stable` and `oldstable` tags which today resolve to 1.22.0 and 1.21.7, but will float as Go versions change.
- Some anonymous function names seem to have changed in Go 1.22 and a test had to be modified as a result

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,14 +12,14 @@ jobs:
     steps:
     - uses: actions/setup-go@v5
       with:
-        go-version: '1.20'
+        go-version: 'oldstable'
     - uses: actions/checkout@v4
     - run: go mod tidy && git diff --exit-code go.mod go.sum
   build:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -12,14 +12,14 @@ jobs:
...
-        go-version: '1.20'
+        go-version: 'oldstable'
     - uses: actions/checkout@v4
     - run: go mod tidy && git diff --exit-code go.mod go.sum
   build:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        version: [ '1.20', '1.21' ]
+        version: [ 'oldstable', 'stable' ]
```
onsi, line 22: :tada: I love this - thank you!

onsi:
thanks @blgm this looks good to me.  perhaps we can avoid relying on internals for the specific test you had to fix - but, honestly, i don't know that it's worth the effort.  appreciate you putting this together!

PR #1357, opened 2024-02-08
core_dsl: disable Getwd() with environment variable

### LINKED ISSUE #1355: Getwd() breaks test caching on build runners like Buildkite
I work on a project that has a test suite that calls ginkgo.RunSpecs. RunSpecs calls os.Getwd() in order to set the SuitePath, which is used mainly for cosmetic purposes - the display in the default reporter. 

Unfortunately os.Getwd() calls os.Getenv("PWD"), which can change from run to run if you are using a test suite runner like e.g. Buildkite. Because test caching relies on environment variables being the same from run to run, this facile change breaks test caching.

It would be great if there was an option to disable the os.Getwd() call, so we can preserve test caching, otherwise we may try to pivot away from using this library altogether.

kevinburke:
os.Getwd() calls os.Getenv("PWD"), which can change from run to run if you are using a test suite runner like e.g. Buildkite. Because test caching relies on environment variables being the same from run to run, this facile change breaks test caching.

Fixes #1355.

### DIFF
--- a/core_dsl.go
+++ b/core_dsl.go
@@ -292,7 +292,7 @@ func RunSpecs(t GinkgoTestingT, description string, args ...interface{}) bool {
 
 	err = global.Suite.BuildTree()
 	exitIfErr(err)
-	suitePath, err := os.Getwd()
+	suitePath, err := getwd()
 	exitIfErr(err)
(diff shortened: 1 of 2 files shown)


onsi:
sorry for the delay - thanks for this!

PR #1365, opened 2024-03-03
docs/index.md: Typo

isodude:
Since it's just a minor nit I don't see any point in opening an issue.

Code was incorrect, this commit fixes that.

Also change the name of the chan variable to be the same as in the invalid case above it.


onsi:
thank you!!

PR #1373, opened 2024-03-13
add `--github-output` for nicer output in github actions

### LINKED ISSUE #1372: Add option to StandardReporter to display things better on github actions
We run some fairly big and verbose test suites on github actions (.e.g: https://github.com/kumahq/kuma/actions/runs/8245083767/job/22550558820).

We could use `::group::` to group the stdout output for example which would make things much easier to read see for docs: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines

That would be an example:
image

This would make the output a lot easier to read.

From a quick look at the code it should likely be somewhere there: https://github.com/onsi/ginkgo/blob/977bc6f813af032b7499f7876a963e6d3f20b685/reporters/default_reporter.go#L284-L306

It could either be a command line flag:  `--github-output` or detect it using en env var like: `GITHUB_ACTIONS=true`.

lahabana:
Leverage github's special formats to output nicer output

https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#using-workflow-commands-to-access-toolkit-functions

Fix #1372

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,4 +27,4 @@ jobs:
         go-version: ${{ matrix.version }}
     - uses: actions/checkout@v4
     - run: go vet ./...
-    - run: go run ./ginkgo -r -randomize-all -randomize-suites -race -trace -procs=2 -poll-progress-after=10s -poll-progress-interval=10s
+    - run: go run ./ginkgo --github-output -r -randomize-all -randomize-suites -race -trace -procs=2 -poll-progress-after=10s -poll-progress-interval=10s
--- a/docs/index.md
+++ b/docs/index.md
@@ -3382,6 +3382,7 @@ When you [filter specs](#filtering-specs) using Ginkgo's various filtering mecha
 Here are a grab bag of other settings:
 
 You can disable Ginkgo's color output by running `ginkgo --no-color`.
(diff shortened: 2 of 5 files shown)


lahabana:
@onsi putting that in draft I still need to validate a few things but that's the rough idea for the change

onsi:
hey that looks great.   the tests for the reporter are a bit long/verbose but will you be adding one as well?

also can you update Ginkgo's own [test.yml](https://github.com/onsi/ginkgo/blob/master/.github/workflows/test.yml) to use the feature?  Thanks!

lahabana:
@onsi I added 1 test (they are indeed quite tricky to write...) let me know if you want more tests

onsi:
hey thanks.  no the one is good - just want to make sure future work doesn't break this feature.

lahabana:
Ok I'll undraft this once I've checked the output in: https://github.com/kumahq/kuma/pull/9603 looks good

lahabana:
@onsi I think it's ready!

onsi:
great!  i’ll merge it in when CI finishes and will cut a release soon.

onsi:
done! thank you :)

PR #1393, opened 2024-04-22
Fix test for gomega version bump

Ygnas:
Gomega error message is different now: `fail(fmt.Sprintf("Context was cancelled (cause: %s)", err))`
This should unblock #1390 

If it would be easier I can bump the version of gomega here as well.
# Verification
- Upgrade gomega to v1.33.0
- All test should pass

### DIFF
--- a/internal/internal_integration/interrupt_and_timeout_test.go
+++ b/internal/internal_integration/interrupt_and_timeout_test.go
@@ -924,7 +924,7 @@ var _ = Describe("Interrupts and Timeouts", func() {
 			Ω(rt).Should(HaveTracked("A"))
 			Ω(reporter.Did.Find("A")).Should(HaveTimedOut(clLine(-1)))
 			Ω(reporter.Did.Find("A")).Should(HaveTimedOut(`A spec timeout occurred`))
-			Ω(reporter.Did.Find("A").Failure.AdditionalFailure).Should(HaveFailed(MatchRegexp("A spec timeout occurred and then the following failure was recorded in the timedout node before it exited:\nContext was cancelled after .*\nExpected\n    <string>: foo\nto equal\n    <string>: bar"), clLine(1)))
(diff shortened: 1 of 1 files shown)


Ygnas:
I'm check it in a sec.

internal/internal_integration/interrupt_and_timeout_test.go
```diff
@@ -924,7 +924,7 @@ var _ = Describe("Interrupts and Timeouts", func() {
 			Ω(rt).Should(HaveTracked("A"))
 			Ω(reporter.Did.Find("A")).Should(HaveTimedOut(clLine(-1)))
 			Ω(reporter.Did.Find("A")).Should(HaveTimedOut(`A spec timeout occurred`))
-			Ω(reporter.Did.Find("A").Failure.AdditionalFailure).Should(HaveFailed(MatchRegexp("A spec timeout occurred and then the following failure was recorded in the timedout node before it exited:\nContext was cancelled after .*\nExpected\n    <string>: foo\nto equal\n    <string>: bar"), clLine(1)))
+			Ω(reporter.Did.Find("A").Failure.AdditionalFailure).Should(HaveFailed(MatchRegexp("A spec timeout occurred and then the following failure was recorded in the timedout node before it exited:\nContext was cancelled .* after .*\nExpected\n    <string>: foo\nto equal\n    <string>: bar"), clLine(1)))
```
blgm, line 927:
```suggestion
                        Ω(reporter.Did.Find("A").Failure.AdditionalFailure).Should(HaveFailed(MatchRegexp("A spec timeout occurred and then the following failure was recorded in the timedout node before it exited:\nContext was cancelled \\(cause: spec timeout occurred\\) after .*\nExpected\n    <string>: foo\nto equal\n    <string>: bar"), clLine(1)))
```
I think the issue is that the `MatchRegexp()` matcher is being used, and the `(` and `)` are tokens in a regular expression, so they need to be escaped.

onsi:
thanks y’all i’ve pulled this in now

Ygnas:
After merge #1381, #1382, #1383 should pass as well, so test could be run there @onsi

blgm:
#1381, #1382, and #1383 are all done. Thanks for fixing the test @Ygnas.

Issue #1401, opened 2024-05-01
when watching, add ability to ignore hidden files


ewollesen:
When watching for changes, I find that ginkgo will detect a change and run the test suite when a new test file is created, even if that test file doesn't match the watch regexp.

For example, if I have a this project layout:

    foo.go
    foo_test.go

...and then I run:

    gingko watch -r --watch-regexp='^[^.].*\.go$'

...then when I create a file ".#foo_test.go", the suite will be run.

I believe this is because it matches `goTestRegExp` defined at https://github.com/onsi/ginkgo/blob/master/ginkgo/watch/package_hash.go#L11, which is evaluated at https://github.com/onsi/ginkgo/blob/master/ginkgo/watch/package_hash.go#L82 and found to have changes.

So the watch regexp isn't able to prevent this behavior.

I've worked up two different ways to work around this behavior, and would be glad to present a PR for either, if this seems like a reasonable ask.

My first (simpler) solution, is just to basically move line 90 in package_has.go to line 81, which effectively makes the watch regexp apply to test files as well as code files. I'm not sure if that's desireable behavior though. I can't think of any cases where a test file wouldn't be matched by the watch regexp, but I'm sure I haven't thought about this as much as the authors.

For a more involved solution, I have a branch that adds an "--ignore-hidden-files" flag to the watch command which just ignores any file with a name that begins with ".". Not sure how this could/should/would be handled for Windows filesystems though..?

Why is this something I even care about? My editor (Emacs), creates a temporary backup file whenever I modify a file. As a result, whenever I modify a test file, my test suite is triggered, which is a waste of time and resources, and can trick me into thinking the suite has passed when it hasn't.

onsi:
hey @ewollesen this is great, thank you!

> but I'm sure I haven't thought about this as much as the authors

I can assure you that, at this point, you absolutely have!

the issue and use-case all make sense to me.  I think adding `--ignore-hidden-files` makes sense though I'm wondering if this should be the default behavior going forward.  What if instead we add `--include-hidden-files` as a configuration option in case folks want to monitor hidden files but, other than that, `ginkgo watch` by default should ignore hidden files.

Thoughts?  A PR would be great!

ewollesen:
> Thoughts? A PR would be great!

[...]

PR #1405, opened 2024-05-06
ignore hidden files

ewollesen:
Hidden files are defined here as those beginning with '.' or '_'.

This is a departure from previous behavior, where hidden source and test files were evaluated via watch. Evaluating hidden files is undesirable because some editors create hidden temporary files, and these files shouldn't trigger test runs.

No flag is added to restore the original behavior at this time as the new behavior is the same as "go test", and it is not expected that it will cause problems. If this change affects you, consider opening a PR to add a CLI flag to restore the previous behavior.

https://github.com/onsi/ginkgo/issues/1401

### DIFF
--- a/ginkgo/watch/package_hash.go
+++ b/ginkgo/watch/package_hash.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"regexp"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


onsi:
This looks great!  I’ll pull it in and cut a release once CI goes green.

PR #1411, opened 2024-05-09
add AddSpecialSuiteFailureReasons to reduce boilerplate

grosser:
basically
```
report.SpecialSuiteFailureReasons = append(report.SpecialSuiteFailureReasons, suite.CompilationError.Error())
->
report.AddSpecialSuiteFailureReasons(suite.CompilationError.Error())
```
because the boilerplate was a bit annoying to see everywhere and now the lines are easier to read

### DIFF
--- a/ginkgo/internal/profiles_and_reports.go
+++ b/ginkgo/internal/profiles_and_reports.go
@@ -107,13 +107,13 @@ func FinalizeProfilesAndReportsForSuites(suites TestSuites, cliConfig types.CLIC
 		}
 		switch suite.State {
 		case TestSuiteStateFailedToCompile:
-			report.SpecialSuiteFailureReasons = append(report.SpecialSuiteFailureReasons, suite.CompilationError.Error())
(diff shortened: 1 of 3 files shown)


onsi:
heyo - i appreciate the boiler plate is annoying but I'd prefer to keep it as-is for now.  the `report` isn't a pointer and i'm trying  to be more explicit about when and how it changes.

grosser:
FYI I started with keeping just 1 method as pointer and that felt weird :)
I like the idea of making it a pointer since I don't like dealing with some things being a pointer and others not, but yeah if you have other plans 🤷 
... feel free to close if you don't think this is mergeable in the near future

onsi:
yeah i hear you.  a lot of the mess in v1 stemmed form me using pointers everywhere and not managing state mutation more carefully.  v2 tried to take a more immutable pov (though, ultimately, there are some pointers here or there).  it definitely helped avoid some of the issues i hit with v1 though it's generated it's own kinds of messes along the way :shrug:

PR #1426, opened 2024-06-09
[build] Allow custom name for binaries.

nunnatsa:
Added the go build flag `-o` for custom path and name of the binary output of `ginkgo build`.


onsi:
lgtm thanks!

PR #1433, opened 2024-07-03
docs: update supported platforms for race conditions

PG2000:
as asked in https://github.com/onsi/ginkgo/discussions/1432


onsi:
thank you!

PR #1463, opened 2024-09-26
Fix the document by replacing `SpecsThatWillBeRun` with `SpecsThatWillRun`

MengjiaLiang:
Hi there,
When I read the Ginkgo doc, I find `SpecsThatWillBeRun` which is mentioned in the doc is not available. The correct name of it is `SpecsThatWillRun` instead.

This chang tries to fix it. Thank you!


onsi:
thank you!

PR #1464, opened 2024-10-07
Allow disabling color with flag GINGKO_COLOR

siretart:
When integrating golang packages using ginkgo, such as CI/CD systems or Linux distributions, there are situations where the test output is tracked in text files and presented without ANSI color decoding.  For those situations, in particular when the tests are invoked directly by `go test`,  and thus without the `gingko` CLI driver, controller color output via an environment variable can be much more convenient

### DIFF
--- a/core_dsl.go
+++ b/core_dsl.go
@@ -20,6 +20,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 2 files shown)


onsi:
they thanks for the PR but this implementation isn't quite right.  i've added support for a `GINKGO_NO_COLOR` environment variable and will cut a new release oon.

siretart:
thanks so much, I'll make sure to update the Debian package so that we can drop my hack

PR #1484, opened 2024-11-11
Add label to serial nodes

### LINKED ISSUE #1483: Add ability to filter for tested marked with "Serial"
Today, "Serial" is not a real label and thus it is impossible to filter by.
It would be great if we could grab serial tests by using this label instead of adding one of our own

akalenyu:
Adding a label in order to be able to filter tests decorated with `Serial`.

Fixes #1483

### DIFF
--- a/integration/timeline_test.go
+++ b/integration/timeline_test.go
@@ -138,7 +138,7 @@ var _ = Describe("Timeline output", func() {
 
 			Ω(session).Should(gbytes.Say(`3 specs`))
 
-			Ω(session).Should(gbytes.Say(`a full timeline\n`))
+			Ω(session).Should(gbytes.Say(`a full timeline \[Serial\]\n`))
 			Ω(session).Should(gbytes.Say(`a flaky test\n`))
 			Ω(session).Should(gbytes.Say(`retries a few times\n`))
 			Ω(session).Should(gbytes.Say(`> Enter \[BeforeEach\] a flaky test`))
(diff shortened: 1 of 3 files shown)


akalenyu:
@onsi Are the failures caused by the PR?

onsi:
yes - this line:

https://github.com/onsi/ginkgo/blob/master/integration/timeline_test.go#L141

needs to changed to `a full timeline [Serial]\n`

you should be able to `ginkgo -r -p` from the root directory and see it all go green.  Every now and then an integration spec will flake on a slow machine but they are generally pretty reliable so if you see a failure it’s _probably_ the new code.  In this case adding `Serial` as a label changed the output and we’ll need to update the test.

akalenyu:
> yes - this line:
> 
> https://github.com/onsi/ginkgo/blob/master/integration/timeline_test.go#L141
> 
> needs to changed to `a full timeline [Serial]\n`
> 
> you should be able to `ginkgo -r -p` from the root directory and see it all go green. Every now and then an integration spec will flake on a slow machine but they are generally pretty reliable so if you see a failure it’s _probably_ the new code. In this case adding `Serial` as a label changed the output and we’ll need to update the test.

Added. I couldn't understand who sets the serial decorator on this timeline test though

onsi:
yay! that worked.  will cut a release soon

akalenyu:
> yay! that worked. will cut a release soon

hey, are you planning to cut a release?

onsi:
done!

Issue #1489, opened 2024-12-05
CSV data are not properly escaped by `ginkgo outline`


belden:
I'm new to Ginkgo as a test framework, and was using `ginkgo outline --format indent | xsv table` to explore the structure of some of our existing tests.¹

I ran into a test which looks like this:

```go
package integration_test

import (
	. "github.com/onsi/ginkgo/v2"
)

var _ = Describe("GET /defaults", Label("integration"), func() {
	When("requesting global, default settings", func() {
		// ... stuff here
	})
})
```

When `ginkgo outline <the-file>` is run on the above contents, the results are these:

```
Name,Text,Start,End,Spec,Focused,Pending,Labels
Describe,GET /defaults,77,214,false,false,false,"integration"
When,requesting global, default settings,135,211,false,false,false,""
```

This output from `ginkgo` is malformed CSV because the phrase `requesting global, default settings` is a single string within the test file, but is presented unquoted in the output.

The proper output is

```
Name,Text,Start,End,Spec,Focused,Pending,Labels
Describe,GET /defaults,77,214,false,false,false,integration
When,"requesting global, default settings",135,211,false,false,false,
```

Notice too that the trailing `""` simply becomes an empty field; `""` in CSV is the proper way to encode a single `"`, similar to how `\"` is the encoding for a `"` in JSON.

I have a pull request which fixes the bug mentioned above. I'm opening this issue in order to reference it from my pull request.

----

¹ The actual command I was running was

```sh
$ fzf \
  --tac \
  --cycle \
  --multi \
  --preview-window=right:75%:wrap \
  --preview $'ginkgo outline --format indent {} | xsv table'
```

which nicely shows the "skeleton" of tests.

PR #1490, opened 2024-12-05
Fix CSV formatting for 'ginkgo outline'

belden:
Ensure CSV formatting properly encodes various edge-case values.
    
We lean on `encoding/csv` from the go stdlib to do the work for
us. In order to mix writing indentation with writing csv rows, we must
csvWriter.Flush() after every csvWriter.Write(row).

Adds a new test file, and updates existing test .csv files to expect
proper CSV encoding. The commit messages on each commit may be useful
to the reviewer.

This pull request resolves https://github.com/onsi/ginkgo/issues/1489


ginkgo/outline/_testdata/position_test.go.csv
```diff
@@ -1,3 +1,3 @@
 Name,Text,Start,End,Spec,Focused,Pending,Labels
-Describe,104,240,104,240,false,false,false,""
-It,209,236,209,236,true,false,false,""
+Describe,"104,240",104,240,false,false,false,
+It,"209,236",209,236,true,false,false,
```
belden, line 3: I laughed to discover that `ginkgo/outline/_testdata/position_test.go` already has text in place which should have forced using a CSV writer. The test in question:

https://github.com/onsi/ginkgo/blob/9e2f337b10cf4973bbc9ae4aeac5a3a334c3571a/ginkgo/outline/_testdata/position_test.go#L1-L22

A test which uses a csv Reader to load each `_testdata/*_test.go.csv` would have caught this, but it's easy to miss out on writing meta-tests.

In order not to clutter this PR, I have not added such a test to the suite. If the reviewers like, I can add such a commit onto this branch; or provide a separate PR.

belden: Here's a picture to hopefully illustrate the previous error in this file:

image

Counting the comma-separated fields in each line, you'll see:

|Comment|Row|Fields|
|-|-|-|
|Header|`Name,Text,Start,End,Spec,Focused,Pending,Labels`|8|
|Removed line|`Describe,104,240,104,240,false,false,false,""`|9|
|Removed line|`It,209,236,209,236,true,false,false,""`|9|
|Added line|`Describe,"104,240",104,240,false,false,false,`|8|
|Added line|`It,"209,236",209,236,true,false,false,`|8|

onsi: ha! this is great, thanks. yes please feel free to add that test to this PR.

belden: Done!

I wrote the test and saw it pass on the HEAD of my branch. I rebased it to pretend I'd written this meta-test as the first, rather than last, commit on this branch.

Then I just checked out each commit and ran

     belden@laptop github/onsi/ginkgo/ginkgo/outline $ go test .

to see how the test failed. With the meta-test as the first commit, it properly fails due to `outline/_testdata/position_test.go.csv` having the extra fields, as shewn in the screenshot above.

The test continues to fail in various ways for the subsequent commits; and when we finally get to the HEAD of this branch, the test passes as expected.

----

I broke the test up into two functions so as not to make the `DescribeTable` test body be too long. When I initially looked at that file, I immediately saw the super helpful comment about how to extend the test. I wanted to let the next person immediately see that same helpful comment.

onsi:
Thank you!  I'll cut a new release soon.

belden:
Ah nerds! Sorry about that @onsi - I saw those files show up, and thought they looked weird. I'm sorry for the extra work for you here.

Here's an approach that works and doesn't tie you to getting a new PR, or otherwise needing to replay the work:

----

1. Make a fresh clone of your repo

    ```sh
    cd /tmp/
    git clone https://github.com/onsi/ginkgo temp-ginkgo
    ```

1. Hop into that repo and pretend we never merged my PR

    ```sh
    cd /tmp/temp-ginkgo
    git reset --hard v2.22.0
    ```

1. Make a new branch to work on. We'll use `origin/master` as the branch point, since that still contains the work we're interested in.

    ```sh
    git checkout -b temp-branch origin/master
    ```

1. Find and remove the `*.test` files. Note that the trailing `+` on the next line is part of the `find` command.

    ```sh
    find . -name '*.test' -exec rm {} +
    ```
    
1. Commit the removal, with a special commit message. Note the sha here is the one you identified.

    ```sh
    git commit -m 'fixup! 3d0e8f1'
    ```

1. Tell `git rebase` to cuddle removing the `*.test` files back onto the offending commit. We need to set some variables to ensure your editor doesn't needlessly have you edit commit messages.

    ```sh
    GIT_EDITOR=/bin/true EDITOR=/bin/true git rebase -i --autosquash --autostash 3d0e8f1^
    ```

----

Now check the `temp-branch`. Some useful checks here:

1. Validate the tagged sha for `v2.22.0` is still `9e2f337b10cf4973bbc9ae4aeac5a3a334c3571a`

    ```sh
    git log --pretty=oneline --n 5 --decorate=full
    ```

1. Validate `*.test` are not present on the `temp-branch` history:

    ```sh
    git diff --name-only v2.22.0... | grep '\.test'
    ```

----

If the history of `temp-branch` is to your satisfaction, then you can merge it into your local `master`:

```sh
git merge temp-branch
```

And then you're a force-push away from wiping out this embarrassing mistake from me

```sh
git push --force-with-lease
```

----

I tried a few different approaches for fixing this, and the above one happened to be the easiest thing that worked. `git rebase --exec $'...'` might be able to work, but it eluded me.
Likewise `git filter-branch`.

This is trivial to do with `git-filter-repo` **but** that tool operates across the entire history of the repo, and rewrites all the SHA1s. Invalidating all of your tags seemed like it would be a no-go.

The automatic fixup trick is one that I use a lot, though behind the veil of `magit.el`. And it's easy enough to check each step, so it's what I've presented here.

onsi:
thanks for taking the time to write this all up.  i'll give it a try!

onsi:
that worked!  fantastic - this was super helpful and clear and I learned a lot more about some serious git-fu.  think you!

Issue #1510, opened 2025-01-21
Ginkgo build reports "Compiled path/suite1/test.test" for each suite


rjacobs31:
## The command being run
```bash
ginkgo build path/suite1/ path/suite2/ path/suite3/
```

## Expected output
```
Compiled path/suite1/test.test
Compiled path/suite2/test.test
Compiled path/suite3/test.test
```

## Actual output
```
Compiled path/suite1/test.test
Compiled path/suite1/test.test
Compiled path/suite1/test.test
```

## Probable cause
Code in `ginkgo/build/build_command.go` is setting `goFlagsConfig.O` on the first completed suite, then re-using that same value when reporting each subsequent suite which was compiled.

It is likely that this does not occur when `-o outputdir/` is passed to `ginkgo build`.

I am more than willing to code a fix, since it does not seem complicated.

rjacobs31:
Apologies. I didn't see the mentioned PR had been merged. I can confirm that this issue has been resolved.

PR #1511, opened 2025-01-21
Fix binary paths when precompiling multiple suites.

rjacobs31:
Fixes issue #1510.

When precompiling multiple suites with one invocation of `go build`, it currently takes the name of the first output binary and reports that as the output binary for every suite after that as well. When `-o outputdir` is passed to `ginkgo build`, this doesn't happen.

This PR doesn't use `goFlagsConfig.O` unless it was actually set beforehand, which prevents the suites from being reported incorrectly.


onsi:
gah sorry for dropping this on the floor - thanks for fixing it!

PR #1512, opened 2025-01-22
docs: fix test workflow badge

dalryan:
It looks like the path to the test workflow badge is not correct, causing the badge to not render correctly. 

It currently shows `no status` instead of `passing`.

This change alters the path to point directly to the workflow run on the master branch.


blgm (commented):
Thank you @dalryan

PR #1517, opened 2025-02-16
Fix: Correct Markdown list rendering in MIGRATING_TO_V2.md

arosh:
I noticed that the following Markdown section doesn't seem to be rendering as a list, even though I believe it's intended to be. As you can see in the screenshot below, the items are not being displayed as a list.

スクリーンショット 2025-02-16 175607
スクリーンショット 2025-02-16 175701


onsi:
thanks

PR #1528, opened 2025-03-13
Fix typo in documentation: DescribeHandleSubtree -> DescribeTableSubtree

arosh:
There is a mention of `DescribeHandleSubtree`, but this name doesn't appear anywhere else in the docs or in the code files. From reading the text around it, I think it should be `DescribeTableSubtree` instead.

Sorry if I misunderstood!


onsi:
thanks!

Issue #1542, opened 2025-04-02
Ginkgo number of Procs on Kuberenetes


emirot:
## Issue

I m running test via Ginkgo and those are executed in K8s via CI.

Ginkgo does not utilize [Automaxprocs](https://github.com/uber-go/automaxprocs)  so if I don't specify `--procs=<number of cpu of the pod>` it will lead to unexpected behavior.

For example when using a c5.12.xlarge it will run across  46 processes instead of the 2 CPU core specified in my pod spec. Which could lead in unwanted behaviors e.g performance degradation, creating spikes if those tests use an API etc.

The expected behavior would have been to run across 2 CPU means 2 ginkgo procs. 

I believe a good amount of people are running Ginkgo in container env and it will make tuning automatic for a them.

PR #1543, opened 2025-04-02
Add automaxprocs to automatically match the linux container CPU Quota

emirot:
## What

Resolve issue https://github.com/onsi/ginkgo/issues/1542


onsi:
hey - sorry for the confusion here.  I pulled this in at first thinking it would be enough but... did you check that this actually fixes the underlying issue in a linux container?  My local machine is a mac and I haven't confirmed that this'll work.  Moreover, Ginkgo uses `runtime.NumCPU()` to compute the number of CPUs - which.. I don't _think_ is governed by `GOMAXPROCS`?

Can you confirm that pulling in `automaxprocs` this is working as you expect in a container?  If not, you'll also need to adjust [these two functions](https://github.com/onsi/ginkgo/blob/master/types/config.go#L155-L176)

emirot:
@onsi Thanks for the feedback you were right I updated here is the new PR https://github.com/onsi/ginkgo/pull/1545

PR #1545, opened 2025-04-06
Add automaxprocs for using CPUQuota

emirot:
## What

Adding automaxprocs so ginkgo can set better CPU in container environments.

Updating to `runtime.GOMAXPROCS(-1)`

As per [documentation](https://pkg.go.dev/runtime#GOMAXPROCS)

```
It defaults to the value of [runtime.NumCPU](https://pkg.go.dev/runtime#NumCPU). If n < 1,
```

Issue: https://github.com/onsi/ginkgo/issues/1542

## Why

Currently ginkgo will take the num CPU  of the host instead of what it is running on.
It would lead to degraded performances and unwanted behaviors.

## Tests

Running on a `e2-custom-16-20480` 16 vcpu

Running within a pod with that spec

```
apiVersion: v1
kind: Pod
metadata:
  name: debian-pod-10-cpu
spec:
  containers:
  - name: debian-container
    image: debian:latest
    resources:
      requests:
        cpu: "10"
        memory: "5Gi" # Optional, but recommended to also specify memory
        ephemeral-storage: "10Gi"
      limits:
        cpu: "10"
        memory: "5Gi" # Optional, but recommended to also specify memory
    command: ["sleep", "infinity"] 
```

Before change

```
======================================================
Random Seed: 1743966102

Will run 183 of 183 specs
Running in parallel across 15 processes
```

After change

```
root@debian-pod-10-cpu:/ginkgo# ./gink -p ./integration/
2025/04/06 18:57:55 maxprocs: Updating GOMAXPROCS=10: determined from CPU quota
Running Suite: Integration Suite - /ginkgo/integration
[integration] 
======================================================
Random Seed: 1743965875

Will run 183 of 183 specs
Running in parallel across 9 processes
```

And also if no cpu quota found gomaxprocs won't be set.


onsi:
looks good thanks!  i'll cut a release soon.

PR #1553, opened 2025-05-02
Fix mangled syntax for JustBeforeEach warnings

pburkholder:
Your documentation is so good that I just had to fix it when I ran into an editing error.


onsi:
❤️ thanks!!

PR #1566, opened 2025-07-04
test(perf): rename "_fixture" to "testdata"

blgm:
The CodeQL tool is complaining:
```
Unable to extract some Go projects
The following 1 Go project could not be extracted successfully:
ginkgo/performance/\_fixtures/performance\_fixture
```

Looks like some corruption around files starting with "_". By renaming "_fixtures" to "testdata" we achieve the aim of being ignored by the Go toolchain while having a friendlier name.


onsi:
hey @blgm - CI went red on this change.  do you mind taking a look when you've got some time?

PR #1567, opened 2025-07-04
chore: fix some CodeQL warnings

blgm:
CodeQL has flagged some warnings:
1) an integer conversion issue in progress_report.go
2) regexp without anchor in dependencies.go

(1) should be fixed by changing the call to ParseInt(), and (2) can be implemented more simply by using substring matching, which will likely make the warning go away.


onsi:
thanks!

Issue #1563, opened 2025-06-19
product version filters


Icarus9913:
I'd like to use ginkgo to test a product with matrix versions. However, there are some features just supported in higher versions of my product.
For example, I use ENV to specify my program, which passes to my logical codes like `PROGRAM_VERSION=1.1 ginkgo ./test/...` and `PROGRAM_VERSION=1.0 ginkgo ./test/...` to test, and some `It` cases just works for `PROGRAM_VERSION=1.1`. 

Surely I can write version verification at the head of the `It` to **Skip** running in 1.0 version , I'd like to know if the current ginkgo already implements such a feature, like using `Labels` to do that ?

lahabana:
One way to go about this would be to use this for runtime: https://onsi.github.io/ginkgo/#skipping-specs

But what I think we would prefer it to be able to Label a test with its minimum version.
I imagine an API similar to [labelSets](https://onsi.github.io/ginkgo/#label-sets) but for semver.

ginkgo --label-filter="integration && !slow && Version: isSemVer ^1". Would this make sense?

onsi:
Hey there sorry for the delay.  There isn't formal support for denver in the label filter language.  If either of you would like to add it in a PR I'd be happy to pull it in.  However I don't have the capacity to take the work on myself now.

Let me know if you'd like to do it and i can point you at the right parts of the code.

Icarus9913:
@onsi Hi, I think I can have a try with it :)

onsi:
sounds good @Icarus9913 the parser is [here](https://github.com/onsi/ginkgo/blob/master/types/label_filter.go) and the tests are [here](https://github.com/onsi/ginkgo/blob/master/types/label_filter_test.go) (I think we can skip an integration test for this as the unit tests are fairly comprehensive).

You'll also want to update the docs [here](https://github.com/onsi/ginkgo/blob/master/docs/index.md) (probably in its own section after [Label Sets](https://github.com/onsi/ginkgo/blob/master/docs/index.md#label-sets)).

In terms of implementation, my instinct is that `Label("1.2.3")`, `Label("v1.2.3")` and `Label("V1.2.3")` could all be interpreted as semver when passed to a semver filter. 

On the syntax size, for the lable filter, the easiest approach might be to introduce a new `isSemVer(X)` wehre `X` would be a valid semver constraint.  A library like [semver](https://github.com/Masterminds/semver) could then interpret the constraint for us so we don't have to deal with all the edge cases!

[...]

PR #1570, opened 2025-07-14
feat: add semantic version filtering API

Icarus9913:
xref: #1563 

This PR aims to implement a new feature to support semantic version filtering. You can use the commands to run specific versioned test cases:

`ginkgo run --sem-ver-filter="2.1.1"`

```text
Describe("Feature with version requirements", func() {
  It("should work in version 3.2.0 and above", SemVerConstraint(">= 3.2.0"), func() {
    // This test will only run when the version constraint is satisfied
  })

  It("should work in a specific version range [2.1.0, 2.3.0)", SemVerConstraint(">= 2.1.0, < 2.3.0"), func() {
    // This test will only run when version is between 2.1.0 (inclusive) and 2.3.0 (exclusive)
  })
})
```


lahabana:
xref: #1563

docs/index.md
```diff
@@ -2664,6 +2664,50 @@ Some more examples:
...
+- `minVersion:vX.Y.Z`: Specifies the minimum version required for the test to run.
+- `maxVersion:vA.B.C`: Specifies the maximum version up to which the test should run.
+
+Both labels are optional. If a test lacks a `minVersion` or `maxVersion` label, it can still be filtered based on the presence or absence of these labels.
+
+You can combine `isSemVer(X)` filter expression evaluates whether a label's version satisfies a given semantic version constraint `X`. The constraint is documented in [semver](https://github.com/Masterminds/semver). Here’s how it works:
+
+- `X` is a valid semver constraint, such as `>=1.2.3`, `<=2.2.3`, `^2.1.0`, or `*` (matches any valid semver).
+- If the label (e.g., `minVersion` or `maxVersion`) is present, `isSemVer(X)` checks if its version satisfies `X`.
+- If there's no version label present on the test, `isSemVer(X)` returns `false`.
```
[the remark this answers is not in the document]
Icarus9913: `Label("minVersion:v1.2.3"), Label("maxVersion:v2.2.3")`
`isSemVer(>= 2.1.0)`
--> true     // because the test case contains a range `[v2.1.0, v2.2.3]` that matches the constraint `[v2.1.0, ~)`

And I thought about the following cases:


`Label("minVersion:v2.3.0"), Label("maxVersion:v2.5.0")`
`isSemVer(>=2.2.0, <2.4.0)`
--> true     // because the test case contains a range `[v2.3.0, v2.4.0)` that matches the constraint `[v2.2.0, v2.4.0)`

---

`Label("minVersion:v2.2.0"), Label("maxVersion:v2.4.0")`
`isSemVer(>=2.3.0, <2.5.0)`
--> true     // because the test case contains a range `[v2.3.0, v2.4.0]` that matches the constraint `[v2.3.0, v2.5.0)`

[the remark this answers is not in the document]
Icarus9913: Thought back onsi's original design, which just defines a version in the label like `Label(2.8.1)`. We can have the following commands to serve:

* ginkgo --label-filter "isSemVer(= v2.1.0)"
* ginkgo --label-filter "isSemVer(!= v2.1.0)"
* ginkgo --label-filter "!isSemVer(= 2.1.0)"
* ginkgo --label-filter "isSemVer(>= v2.1.0)"
* ginkgo --label-filter "isSemVer(<= v2.1.0)"
* ginkgo --label-filter "isSemVer(*)"
* ginkgo --label-filter "isSemVer(~2.8.0)" 
* ginkgo --label-filter "isSemVer(^2.8.0)"

Although, the above use cases look good but if you think deeply, you'll noticed that the following use case can make you confused:

For example, if there's one `It("my test", Label("2.1.0"))` and your software version is `2.9.0`, then you need to think about how to write your label filter and it's ambiguous if the `It` case only support `>= 2.1.0, < 2.2.0` or this case is just supported from `2.1.0`. Later if you have **thousands of** different label versions, you'll mess up the `isSemVer` label filter and make it as a burden.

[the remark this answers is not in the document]
Icarus9913: With the above 2 ways(Label version and Label Sets), somehow I feel like both of them are not easy to use once the constraint becomes complicated with the given Label Set version range.

And back to my original goal, what I want actually is `SOFTWARE_VERSION=1.1 ginkgo ./test/...`
We should put the constraint on the container and just pass a software version to ginkgo label filter. So it should look like this:

```text
It("test 1", SemConstraint(">= 2.1.0, < 2.2.0"), func() {
	...
}) 

ginkgo --label-filter "SemVersion: 2.1.1"
```

In this, for the implementation once we have such a label filter `SemVersion` keyword, we check each container if it has the `SemConstraint` keyword. Then let the container's constraint verify if it matches the given `SemVersion`.  I'd like to say this way is maybe easier to understand and implement.  WDYT? @onsi @lahabana 

There's just one thing that I'm not sure about: if all of the nested 'Describe' and 'It' containers have the `SemConstraint`, should it be overwritten or merged?

```text
Describe("Feature with version requirements", SemConstraint(">= 2.1.0, < 2.2.0"), func() {
  It("should only run when v1.2.3 <= version <= v2.0.0", SemConstraint(">= 1.2.3, <= 2.0.0"), func() {
    // This test will only run when the version constraint is satisfied
  })
})
```

docs/index.md
```diff
@@ -2664,6 +2664,50 @@ Some more examples:
...
+
+It("min version test", Label("minVersion:v2.0.0"), func() {
+  // Test code here
+})
+```
+
+Tests can be labeled with `minVersion` and `maxVersion` to specify the version range in which they are applicable.
+
+- `minVersion:vX.Y.Z`: Specifies the minimum version required for the test to run.
+- `maxVersion:vA.B.C`: Specifies the maximum version up to which the test should run.
```
Icarus9913, line 2690: For the implementation, should we just use the hardcoded string `minVersion` and `maxVersion` for parsing or should it still be customized by the user? I'm talking about this because we need to think about an efficient way to parse the version, otherwise parse all of the labels would cause performance abuse.

Icarus9913:
@onsi Hey onsi, I think I finished the basic implementation. For better review, you can read my designs in the xref issue.
One thing need to be mentioned, since SemVer repo supports such constraints
```text
// no errors
c, err := semver.NewConstraint("<= 1.2.3, >= 1.4")
```

I make the constraints combined by the comma symbol to build a new constraint, which solves the hierarchy problem and is easy to combine just like the `Label` behaviors.

Let me know if you agree with this design, then I'll keep going to supplement some tests.

onsi (changes requested):
Hey this is really good - it’s thorough and covers a lot of edges.  Thanks for putting so much effort into it :)

I left a few minor comments that need to be addressed and you should also add an integration test.  But other than that, I think we’re good to go.

Icarus9913:
@onsi Hey, I supplemented the integration tests. Please let me know if I missed something, then I will supplement it in the PR. Or you can merge this PR and let me create a separate PR to finish the missing stuff to avoid making thi PR too large to review :)

onsi:
hey @Icarus9913 - i see you've added some new fixtures and that's the correct first step.  But you also need to add an actual test that exercise those fixtures.  See, for example, this test:

https://github.com/onsi/ginkgo/blob/master/integration/filter_test.go

I think a new test in `filter_test.go` that exercises the fixtures you added would be helpful. Apologies if you did this and I somehow missed this!

Icarus9913:
OK I see, I just tested it locally and found lots of failures. Maybe need some time to fix it.

Icarus9913:
@onsi Hey, I pushed a new commit. Let me know if there's anything missing, and you can also help me improve some description sentences since I'm not a native English speaker :)

docs/Gemfile.lock
```diff
@@ -234,8 +234,7 @@ GEM
     minitest (5.25.5)
     net-http (0.6.0)
       uri
-    nokogiri (1.18.9)
-      mini_portile2 (~> 2.8.2)
+    nokogiri (1.18.9-x86_64-linux-gnu)
```
onsi, line 237: hmm. not sure how this leaked in.  can you revert the lock file?

integration/filter_test.go
```diff
@@ -103,4 +105,53 @@ var _ = Describe("Filter", func() {
...
+				"shouldn't expand the constraint",
+				"should run without constraints by table driven",
+				"should run with version in range [2.0.0, ~) by table driven",
+			}
+			skippedSpecs := []string{
+				"shouldn't run with version in a conflict range",
+				"shouldn't combine with a conflict constraint",
+				"shouldn't run with version in a conflict range by table driven",
+			}
+			Ω(specs).Should(HaveLen(len(passedSpecs) + len(skippedSpecs)))
```
onsi, line 138: i think we need a stronger test here.  You could harvest the leaf node texts and use Gomega’s `ConsistsOf()` to help you match the strings.  Alternatively you could iterate through every spec from the json file and confirm that it’s semver version passes the filter.

onsi (commented):
heyo @Icarus9913 i left a comment.  let me know if you’re up for getting this past the finish line.  thanks for all your work on this!

docs/Gemfile.lock
Icarus9913: reverted the whole lock file

Icarus9913:
@onsi Hey, I pushed a new commit. Thanks for your review.

onsi:
hey this is looking great!  I'm ready to pull it in but it looks like it needs to be rebased on top of the latest master?  Could you do that?  I'll pull it in tomorrow.

onsi:
ginkgo 2.24.0 is out now with the semver filter support!  Thanks @Icarus9913

PR #1573, opened 2025-07-31
fix(types): ignore nameless nodes on FullText()

maxcleme:
We stumbled upon an issue where `.FullText()` was returning a leading space in the Spec name due to an empty `Describe`.

Not entirely sure if this behavior is intentional or not, but in any case, the pull request is ready if you’d like to take a look and consider accepting it 🙂

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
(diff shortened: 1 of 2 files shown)


onsi:
hey sorry for the delay.  thanks for the fix!

PR #1574, opened 2025-07-31
Clarify SynchronizedBeforeSuite behavior for allProcesses and process1 phases

sachinDcoder:
- Clarified that the allProcesses function in SynchronizedBeforeSuite runs on all parallel processes, including process #1.
- These updates help prevent confusion when coordinating setup tasks across parallel test processes.


tfrankland1 (approved):

onsi:
?? the committed changes don't match the description on the issue.

Issue #1107, opened 2022-12-28
Convert json output into test2json output format


stotz89:
Hello, 
I know there are already a couple of issues and some discussions related to that topic however I wanted to ask this question. Since v2 Ginkgo produces its own json format as an output of the test results.

We are using Sonarqube to report back test coverage and test reports and unfortunately, at least as far as I know, Sonar expects the test reports in the test2json format produced by `go test`: 

> Path to test execution reports generated by Go with '-json' key, available since go1.10 (e.g.: go test -json > test-report.out).

One the one hand we could only use `go test` to get the desired outputs for test reports & coverage however on the other hand the output Ginkgo produces is just much more detailed that we also would like to see (in our CI pipeline). Therefore, our mitigation currently is to execute both steps (`go test` & `ginkgo`) to have the best of both but as said, should only be a mitigation.

So, my questions would be: 
1. Someone else have already faced & solved this issue? Could a `ReportAfterSuite` convert the report back into the test2json format?
2. Would it be a feasible that Ginkgo can bring back the support for a test2json output format?

onsi:
hey there @stotz89 

I don't have a lot of experience/context with the `test2json` format though looking through it it does seem to be oriented towards streaming test information vs generating the comprehensive final report.  With that said, I could imagine custom code in `ReportAfterSuite` generating ` test2json` formatted report that probably gets close enough.  I could also imagine a simple post-processing tool that can take a ginkgo generated JSON file and produce a `test2json` formatted file.

> Would it be a feasible that Ginkgo can bring back the support for a test2json output format?

Just to clarify - there never has been support for the `test2json` format.  Ginkgo's JSON format is new to V2 and did not displace/replace anything prior to v2.  You can actually run `go test -json > test-report.out` with a Gingko suite but you'll simply get one test in the report: the test function in the suite's [bootstrap](https://onsi.github.io/ginkgo/#bootstrapping-a-suite).

As for how to add support for this in Ginkgo I think I'd lean twoards something like a new function in the `reporters` package that looks like:

```
func GenerateGoTestJSONReport(report types.Report, filename string)  error {
    // walk report and generate test2json-compatible objects
    // JSON-encode the objects into filename
}
```

Then users who want a `test2json` formatted report could simply add:

```
ReportAfterSuite("test2json", func(report types.Report) {
   Expect( reporters.GenerateGoTestJSONReport(report, "path/to/report.json")).To(Succeed())
})
```

[...]

PR #1585, opened 2025-08-26
add Go JSON reporter

pseudo-su:
The Goal of this PR is to allow the generation of test reports in the test2json format which is the what `go test -json` emits.

The benefits of implementing this as an inbuilt report output format is that it allows it to behave much more consistently with the other [ginkgo report config/options](https://onsi.github.io/ginkgo/#generating-machine-readable-reports) like

- `--keep-separate-reports`
- `--output-dir=<dir>`, 
- reports will be namespaced with the name of the package: PACKAGE_NAME_report.json

Resolves https://github.com/onsi/ginkgo/issues/1107

## Changes

Core changes:

- Add config flag `--gojson-report=<filename>`
- Added a `internal/reporters` package to limit the surface area of what is kept public in `reporters`
  - `GenerateGoTestJSONReport`
  - `MergeAndCleanupGoTestJSONReports`
- Added the `--gojson-report=report.go.json` option to any integration test that also had the `--teamcity-report` option
- Implement the generation of a report following the `test2json` output

Other changes:

- Updated the tests in the `reporters` package to
    - Capture the report output of different types as snapshots using `go-snaps` (gojson, junit, teamcity)
    - gojson: parse the output report using `tparse` and capture a basic overview of the data in the snapshot
    - junit: parse the output report using `go-junit` and capture a basic overview of the report in the snapshot
    - use a static timestamp instead of `time.Now()` for deterministic snapshot output.
    - dependencies added:
        - [github.com/gkampitakis/go-snaps v0.5.14](github.com/gkampitakis/go-snaps) to snapshot test output in tests
        - [github.com/joshdk/go-junit v1.0.0](github.com/joshdk/go-junit) to parse junit output
        - [github.com/mfridman/tparse v0.18.0](github.com/mfridman/tparse) to parse gotest output

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -3716,22 +3716,25 @@ The resulting JSON file encodes an array of `types.Report`.  Each entry in that 
 
 When possible, we recommend building tooling on top of Ginkgo's JSON format and using Ginkgo's `types` package directly to access the suite and spec reports.  The structs in the package include several helper functions to interpret the report.
 
-Ginkgo also supports generating JUnit reports with
+Ginkgo also supports generating output in additional formats, however Ginkgo specs carry much more metadata than can easily be mapped onto different output reports formats, so information may be lost and/or a bit harder to decode than using Ginkgo's native JSON format.
 
 ```bash
+# JUnit report
 ginkgo --junit-report=report.xml
+# Teamcity report
+ginkgo --teamcity-report=report.teamcity
+# Go json report (same format as `go test -json`)
+ginkgo --gojson-report=report.go.json
 ```
 
-The JUnit report is compatible with the JUnit specification, however Ginkgo specs carry much more metadata than can be easily mapped onto the JUnit spec so some information is lost and/or a bit harder to decode than using Ginkgo's native JSON format.  Nonetheless, Ginkgo does its best to populate as much of the JUnit report as possible.  This includes adding additional metadata using [labels](#spec-labels) - in particular if you provide a label of the form `Label("owner:XYZ")`, the generating JUnit spec will set the `Owner` attribute to `XYZ`.
+Ginkgo does it's best to populate relevant fields and attributes across different report formats. This includes adding additional metadata using [labels](#spec-labels), in particular if you provide a label of the form `Label("owner:XYZ")`, the generated JUnit report will set the `Owner` attribute to `XYZ`.
 
-Ginkgo also supports Teamcity reports with `ginkgo --teamcity-report=report.teamcity` though, again, the Teamcity spec makes it difficult to capture all the spec metadata.
-
-All the machine-readable reports include the full `-vv` version of the timeline for all specs.  This allows you to run Ginkgo in CI with the normal verbosity setting but still get all the detailed information in the machine-readable format.
+All the machine-readable reports include the full `-vv` version of the timeline for all specs. This allows you to run Ginkgo in CI with the normal verbosity setting but still get all the detailed information in the machine-readable format.
 
 Of course, you can generate multiple formats simultaneously by passing in multiple flags:
 
 ```bash
-ginkgo --json-report=report.json --junit-report=report.xml
+ginkgo --json-report=report.json --junit-report=report.xml --gojson-report=report.go.json
 ```
 
 By default, when any of these command-line report flags are provided Ginkgo will generate a single report file, per format, at the passed-in file name.  If Ginkgo is running multiple suites (e.g. `ginkgo -r --json-report=report.json`) then _all_ the suite reports will be encoded in the single report file.


onsi:
hey this does not convert into go's format.  what you've got here is equivalent to `ginkgo -json-report=out.json`

if you'd like to convert ginkgo's json format into go's test format i'd recommend  building something that can take ginkgo's JSON as input and spit out to as output.

pseudo-su:
> does not convert into go's format. what you've got here is equivalent to `ginkgo -json-report=out.json`

Yes I know this PR isn't ready, currently working through understand how the flag config for Ginkgo is setup to make this work

> flag provided but not defined: -ginkgo.gojson-report

edit: nvm this was caused by using a different version of ginkgo as the library vs binary

onsi:
ok, thanks.  So one concern I have with going down this road is that Ginkgo's json format is quite rich and mirrors much of Ginkgo's structure and feature set.  The go test2json format is a fair bit thinner and while it's true that tooling has been built on top of it my experience with the junit reporter is that people invariably end up asking for more information to be surfaced in these other formats than can really fit in them.

It's why I prefer for Ginkgo's json format to be the main thing Ginkgo spits out and for other tooling to come online that can take that format and spit out alternative formats and/or views.  And I'd love for folks who have built deep/detailed ginkgo specs and are invested in that ecosystem to build on top of its native json format.

All of this is likely wishful thinking and I know I'll end up gratefully pulling in whatever you land on.  Just want to call out: there are reasons there isn't already support for the go json format.

pseudo-su:
Yeah I understand the concerns about maintaining different formatting outputs. I wanted to take a stab at implementing it before advocating for it too much.

As a soft counterpoint, we have done a similar thing to what you suggested here https://github.com/onsi/ginkgo/issues/1107#issuecomment-1366881749 for a while and it's been somewhat annoying to try and seamlessly plug-in/integrate well with enabling/configuring other [ginkgo report options/settings](https://onsi.github.io/ginkgo/#generating-machine-readable-reports) like

- `--keep-separate-reports`
- `--output-dir=<dir>`, 
- reports will be namespaced with the name of the package: PACKAGE_NAME_report.json

I'm hoping that implementing it as a "proper" native report format/type will make it behave much more consistently

onsi:
all fair points, makes sense.  lemme know if you run into issues/have questions.

pseudo-su:
Okay @onsi, this is probably in a state where I think it's worth considering.

I'm fairly confident w/ the core changes, and I've tried to follow existing examples on what I think some preferences are in terms of handing test names in output and capturing stderr/stdout etc.

I've potentially taken some liberty (eg adding snapshot tests) that I'd totally understand if there was a preference to revert, change or structure differently. Mostly they were to help get a better understanding of how the existing reporters work.

I'll probably add some comments to bits that I think are worth highlighting

internal/reporters/gojson_event_writer.go
```diff
@@ -0,0 +1,111 @@
...
+	if stdOut != "" {
+		events = append(events, &gojsonEvent{
+			Time:        &specReport.o.EndTime,
+			Action:      GoJSONOutput,
+			Test:        specReport.testName,
+			Package:     report.goPkg,
+			Output:      ptr(stdOut),
+			FailedBuild: "",
+		})
+	}
```
pseudo-su, line 89: This is a fairly naive way of outputting all the  stdout and stderr data captured into a single `test2json` event, technicaclly the `test2json` format would support more granually going through the timeline and emitting more fine-grained events (what the default reporter is doing) and emit seperate test2json output events.

Ond of the downsides of capturing it this way this I've made the decision to just associate the output w/ the `spec.EndTime`. processing the timeline manually into `test2json` events would mean that the timestamps of stderr/stdout could be captured correctlly

internal/reporters/gojson.go
```diff
@@ -0,0 +1,158 @@
...
+		if len(labels) > 0 {
+			name = name + " [" + strings.Join(labels, ", ") + "]"
+		}
+		semVerConstraints := spec.SemVerConstraints()
+		if len(semVerConstraints) > 0 {
+			name = name + " [" + strings.Join(semVerConstraints, ", ") + "]"
+		}
+		name = strings.TrimSpace(name)
+		return name
+}
```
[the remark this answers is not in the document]
pseudo-su: One question I have mostly just out of curiosity is that the junit reporter has a bunch of options:

```go
type JunitReportConfig struct {
	// Spec States for which no timeline should be emitted for system-err
	// set this to types.SpecStatePassed|types.SpecStateSkipped|types.SpecStatePending to only match failing specs
	OmitTimelinesForSpecState types.SpecState

	// Enable OmitFailureMessageAttr to prevent failure messages appearing in the "message" attribute of the Failure and Error tags
	OmitFailureMessageAttr bool

	//Enable OmitCapturedStdOutErr to prevent captured stdout/stderr appearing in system-out
	OmitCapturedStdOutErr bool

	// Enable OmitSpecLabels to prevent labels from appearing in the spec name
	OmitSpecLabels bool

	// Enable OmitSpecSemVerConstraints to prevent semantic version constraints from appearing in the spec name
	OmitSpecSemVerConstraints bool

	// Enable OmitLeafNodeType to prevent the spec leaf node type from appearing in the spec name
	OmitLeafNodeType bool

	// Enable OmitSuiteSetupNodes to prevent the creation of testcase entries for setup nodes
	OmitSuiteSetupNodes bool
}
```

But as far as I can tell there doesn't seem to be any way to control/configure those options, so they all default to false (not omitting any behavior)? is there something I'm missing about configurable test report output I should be considering?

internal/reporters/gojson_event_writer.go
onsi: not sure what's best, personally.  one tricky thing to consider is once this is shipped making breaking changes will be difficult if folks start to rely on the format.  In principle emitting more granular information will enable more usecases... in practice, I don't know how much people will actually care/want that in the gotestjson format vs just grabbing it from ginkgo's native format.

pseudo-su: > once this is shipped making breaking changes will be difficult if folks start to rely on the format
In practice I think the risk of this is small enough to be manageable, in practice the way that something would consume the `test2json` format would be to concatenate all the `type: output` events to interpret output for the tests.
That's pretty much how rebuilding the test output works in [this test](https://github.com/onsi/ginkgo/pull/1585/files#diff-cf598164e0e40ed900b62afe1298872a15af5db19c515ae92b7ba64314c4426bR82-R88).

Because of how the stream of `output` events in the `test2json` format is meant to be processed having it in one event vs many might not need to be considered much of a "breaking change" if there was a desire to update it in the future.

> in practice, I don't know how much people will actually care/want that in the gotestjson format

Yeah, I agree. Technically the use-case I have for this doesn't actually require capturing the stderr/stdour at all. We just have internal tooling that is built reading/interpreting test results/counts/failures etc from the `go test -json` output and aggregating that across ginkgo test suites has been fiddly.

pseudo-su:
@onsi have you had a chance to think consider this?

Would something along these lines be acceptable, and/or what changes would you like to see?

onsi:
hey @pseudo-su - sorry for the delay.  I think this looks good - just needs some documentation to the [machine-readable reports section](https://github.com/onsi/ginkgo/blob/master/docs/index.md#generating-machine-readable-reports).  Would you be up for writing some?  Anything else you want to add/adjust before I pul lit in?  I think it would be good to label the feature as "beta" for a few versions to see if we get any feedback that would warrant breaking changes.  No action there - just a thing we say we're gonna do in the release notes.

pseudo-su:
No worries at all, I appreciate you taking the time 

> just needs some documentation to the [machine-readable reports section](https://github.com/onsi/ginkgo/blob/master/docs/index.md#generating-machine-readable-reports). Would you be up for writing some?

Yeah sure, I've made some updates. Let me know if it's what you were thinking

> Anything else you want to add/adjust before I pulll lit in?

I'm happy w/ where it's currently at

> I think it would be good to label the feature as "beta" for a few versions to see if we get any feedback that would warrant breaking changes. No action there - just a thing we say we're gonna do in the release notes.

makes sense, if it's just in the release notes I won't mention that in the reports section doc

onsi:
looks good to me now, thanks! will pull this in once CI goes green then cut a release

onsi:
had to fix a couple of things by hand but all good now.!

PR #1588, opened 2025-08-28
add --github-output group for progress report content

koct9i:
Keep massage and spec nodes visible, fold everything else.


koct9i:
Here is an example how this problematic output looks without such folding:
https://github.com/ytsaurus/ytsaurus-k8s-operator/actions/runs/17273350550/job/49023719428?pr=546#step:9:139

onsi:
👍 this makes sense to me, thanks for the PR.

PR #1596, opened 2025-09-20
Container-aware GOMAXPROCS is available in golang 1.25

dims:
The support for supporting better GOMAXPROCS using  `"go.uber.org/automaxprocs"` was added in:
https://github.com/onsi/ginkgo/pull/1545

this is not needed in golang 1.25 onwards:
https://go.dev/doc/go1.25#container-aware-gomaxprocs

So let's just ensure we use it only for 1.23 and 1.24 (since the base go.mod is 1.23) and skip it for any other golang versions above those.

### DIFF
--- a/ginkgo/main.go
+++ b/ginkgo/main.go
@@ -3,7 +3,6 @@ package main
 import (
 	"fmt"
 	"os"
-	_ "go.uber.org/automaxprocs"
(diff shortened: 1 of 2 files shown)


onsi:
perfect, thanks @dims !

PR #1606, opened 2025-10-19
enhance tree construction

pohly:
This enables custom frameworks to modify and/or check how the test tree is constructed. This consists of two parts:
- Enabling manipulation of the NewNode parameters before a new node gets added to the tree.
- Retrieving information about the current tree.

The use case in Kubernetes for the first point is supporting special arguments (e.g. `framework.WithSlow()`) which inject additional text (`[Slow]`) as well as a `Slow` label. This is currently done with special wrappers, but it is a common pitfall with confusing errors to forget to switch to those wrappers. Doing it consistently could be enforced with the forbidigo linter, but only in Kubernetes and would imply making lots of changes. More importantly, changes are planned where all downstream users also would have to switch (see next point) which is harder to enforce.

The second point enables using `WithSlow` in a Context node and then adding the `[Slow]` text to the It node based on which labels were added earlier. Note that we haven't decided if or when we want to change this (https://github.com/kubernetes/kubernetes/pull/134686#discussion_r2441715071).

Both is being prototyped in https://github.com/kubernetes/kubernetes/pull/134708.

### DIFF
--- a/internal/group.go
+++ b/internal/group.go
@@ -110,6 +110,7 @@ func newGroup(suite *Suite) *group {
 	}
 }
 
+// initialReportForSpec constructs a new SpecReport right before running the spec.
 func (g *group) initialReportForSpec(spec Spec) types.SpecReport {
 	return types.SpecReport{
 		ContainerHierarchyTexts:             spec.Nodes.WithType(types.NodeTypeContainer).Texts(),
@@ -127,6 +128,34 @@ func (g *group) initialReportForSpec(spec Spec) types.SpecReport {
 		IsInOrderedContainer:                !spec.Nodes.FirstNodeMarkedOrdered().IsZero(),
 		MaxFlakeAttempts:                    spec.Nodes.GetMaxFlakeAttempts(),
 		MaxMustPassRepeatedly:               spec.Nodes.GetMaxMustPassRepeatedly(),
+	}
+}
+
+// constructionNodeReportForTreeNode constructs a new SpecReport right before invoking the body
+// of a container node during construction of the full tree.
+func constructionNodeReportForTreeNode(node *TreeNode) *types.ConstructionNodeReport {
+	var report types.ConstructionNodeReport
+	// Walk up the tree and set attributes accordingly.
+	addNodeToReportForNode(&report, node)
+	return &report
+}
+
+// addNodeToReportForNode is conceptually similar to initialReportForSpec and therefore placed here
+// although it doesn't do anything with a group.
+func addNodeToReportForNode(report *types.ConstructionNodeReport, node *TreeNode) {
+	if node.Parent != nil {
+		// First add the parent node, then the current one.
+		addNodeToReportForNode(report, node.Parent)
+	}
+	report.ContainerHierarchyTexts = append(report.ContainerHierarchyTexts, node.Node.Text)
+	report.ContainerHierarchyLocations = append(report.ContainerHierarchyLocations, node.Node.CodeLocation)
+	report.ContainerHierarchyLabels = append(report.ContainerHierarchyLabels, node.Node.Labels)
+	report.ContainerHierarchySemVerConstraints = append(report.ContainerHierarchySemVerConstraints, node.Node.SemVerConstraints)
+	if node.Node.MarkedSerial {
+		report.IsSerial = true
+	}
+	if node.Node.MarkedOrdered {
+		report.IsInOrderedContainer = true
 	}
 }
 
--- a/reporting_dsl.go
+++ b/reporting_dsl.go
@@ -27,11 +27,38 @@ CurrentSpecReport returns information about the current running spec.
 The returned object is a types.SpecReport which includes helper methods
 to make extracting information about the spec easier.
 
+During construction of the test tree the result is empty.
+
 You can learn more about SpecReport here: https://pkg.go.dev/github.com/onsi/ginkgo/types#SpecReport
 You can learn more about CurrentSpecReport() here: https://onsi.github.io/ginkgo/#getting-a-report-for-the-current-spec
 */
 func CurrentSpecReport() SpecReport {
 	return global.Suite.CurrentSpecReport()
+}
+
+/*
+ConstructionNodeReport describes the container nodes during construction of
+the spec tree. It provides a subset of the information that is provided
+by SpecReport at runtime.
+
+It is documented here: [types.ConstructionNodeReport]
+*/
+type ConstructionNodeReport = types.ConstructionNodeReport
+
+/*
+CurrentConstructionNodeReport returns information about the current container nodes
+that are leading to the current path in the spec tree.
+The returned object is a types.ConstructionNodeReport which includes helper methods
+to make extracting information about the spec easier.
+
+May only called during construction of the spec tree. It panics when
+called while tests are running. Use CurrentSpecReport instead in that
+phase.
+
+You can learn more about ConstructionNodeReport here: [types.ConstructionNodeReport]
+*/
+func CurrentTreeConstructionNodeReport() ConstructionNodeReport {
+	return global.Suite.CurrentConstructionNodeReport()
 }
 
 /*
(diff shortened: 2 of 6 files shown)


reporting_dsl.go
```diff
@@ -27,6 +27,10 @@ CurrentSpecReport returns information about the current running spec.
 The returned object is a types.SpecReport which includes helper methods
 to make extracting information about the spec easier.
 
+During construction of the test tree, the returned report provides
+information about the current container node. Fields related to
+a leaf node are not set in that case.
```
pohly: This is up for debate. It's convenient because it reuses much of the existing API (function and type) and infrastructure (global entry in suite), but it's also a bit confusing because it's not perfectly clear what is being set in the result.

internal/group.go
```diff
@@ -130,6 +131,36 @@ func (g *group) initialReportForSpec(spec Spec) types.SpecReport {
 	}
 }
 
+// initialReportForTreeNode constructs a new SpecReport right before invoking the body
+// of a container node during construction of the full tree.
+func initialReportForTreeNode(node *TreeNode) types.SpecReport {
```
pohly: I put this here to have it near `initialReportForSpec`. It could be made more complete if the caller had access to a `group` instance (needed for `ParallelProcess` and `RunningInParallel`), but perhaps that's not needed.

Other fields (`IsSerial`, `IsInOrderedContainer`, `MaxFlakeAttempts`, `MaxMustPassRepeatedly`) probably can be implemented as-is, I just haven't done it yet.

core_dsl.go
```diff
@@ -501,6 +501,58 @@ func pushNode(node internal.Node, errors []error) bool {
...
+//	    origTransformNodeArgs = ginkgo.TransformNodeArgs
+//	    ginkgo.TransformNodeArgs = transformNodeArgs
+//	}
+//
+//	func transformNodeArgs(nodeType types.NodeType, offset ginkgo.Offset, text string, args []any) (string, []any, []error) {
+//	    location := types.NewCodeLocation(offset) // Can be used in errrors about the call site.
+//	    ... replace custom arguments ...
+//	    return origTransformNodeArgs(nodeType, offset+1, text, args)
+//	}
+var TransformNodeArgs = func(nodeType types.NodeType, offset Offset, text string, args []any) (string, []any, []error) {
```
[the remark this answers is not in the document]
onsi: tl;dr - there's a mirror of the (typically `.` imported DSL) in the `dsl` subpackage that allows people to `.` import just the bits they need.   And I have a test that catches if a publicly exported DSL declaration isn't mirrored in one of the `dsl` packages.

but yes, let's ignore this until we pin down the API.

onsi:
hey @pohly I like where this is headed and the usecase and solution makes sense to me.  Thanks for taking the time to put together a PR!

I've got two thoughts about the API design:

---

On the reuse of `CurrentSpecReport()` and the `SpecReport` struct:  Over the years I've seen a fair bit of confusion about the difference between the tree construction phase and the run phase, and I'd like to try to avoid obscuring that distinction further.

I think we should have a new structure: `TreeConstructionNodeReport` that has the subset of `SpecReport` bits that are needed and relevant, and potentially additional pieces and/or helper methods that make more sense for this usecase.

I think we can then either have

```go
var TransformNodeArgs = func(nodeType types.NodeType, offset Offset, text string, args []any, report TreeConstructionNodeReport) (string, []any, []error) {
```

or add a new

```go
CurrentTreeConstructionNodeReport()
```

that errors if its called outside of the tree construction phases.

Thoughts?

---

Overriding a package level variable is OK but puts the onus on the user to manage a stack and to keep track of and call the original function.  An alternative approach would be to provide the user with a mechanism to register (and potentially unregister if we think it's necessary) additional callbacks:


```
type NodeArgsTransformer func(nodeType types.NodeType, offset Offset, text string, args []any) (string, []any, []error))

var globalArayOfTransfomers []NodeArgsTransformer

func PushTreeConstructionNodeArgsTransformer(f NodeArgsTransformer) {
    globalArrayOfTransformers = append(globalArrayOfTransfomers, f)
}

func transformNodeArgs(nodeType types.NodeType, offset Offset, text string, args []any) (string, []any, []error) {
    for _, f := range globalArrayOfTransformers {
        // bail on error, pass transformed args down, etc...
    }
}
```

if we allow deregistration that you could imagine doing per-container stuff like

```
Describe("foo", func() {
    PushTreeConstructionNodeArgsTransformer(blah)
    defer(PopTreeConstructionNodeArgsTransformer())
})
```

no idea if that would be useful at all so YAGNI.  But still I think registering the callback is a better interface than overriding the package variable.

Thoughts?

---

I'd be happy to help with either of these pieces - but let's discuss together before we dive into implementation.

pohly:
> CurrentTreeConstructionNodeReport + PushTreeConstructionNodeArgsTransformer

Both makes sense.

> I'd be happy to help with either of these pieces

You've done enough already for Kubernetes, I can make those changes.

If you have time, then perhaps look into the idea of prioritizing specs during parallel execution?

pohly:
> I can make those changes.

I've pushed an update with the proposed API. The content of NodeConstructionReport is tentative, I only selected fields which seemed to make sense without populating all of them.

core_dsl.go
```diff
@@ -501,6 +501,69 @@ func pushNode(node internal.Node, errors []error) bool {
...
+// A NodeArgsTransformer can be registered with AddTreeConstructionNodeArgsTransformer.
+type NodeArgsTransformer func(nodeType types.NodeType, offset Offset, text string, args []any) (string, []any, []error)
+
+// AddTreeConstructionNodeArgsTransformer registers a NodeArgsTransformer.
+// Only nodes which get created after registering a NodeArgsTransformer
+// are transformed by it. The returned function can be called to
+// unregister the transformer.
+//
+// Both may only be called during the construction phase.
+func AddTreeConstructionNodeArgsTransformer(transformer NodeArgsTransformer) func() {
```
pohly: Instead of Push/Pop I went for Add/remove. That way the ordering of calls doesn't matter and one caller cannot accidentally pop the transformer pushed by another.

I intentionally left out locking because it shouldn't be needed.

pohly: Initially I didn't want to guarantee any ordering (= not documented, iterating over map which forces randomness). That matches the random adding of transformers from different packages.

But for your use case with add/defer remove within a container node's body a deterministic order makes sense. I'm leaning towards "most recent first", as if it was on top of a stack.

core_dsl.go
```diff
@@ -501,6 +502,82 @@ func pushNode(node internal.Node, errors []error) bool {
...
+	transformer NodeArgsTransformer
+}
+
+// transformNewNodeArgs is the helper for DSL functions which handles NodeArgsTransformers.
+//
+// Its return valus are intentionally the same as the internal.NewNode parameters,
+// which makes it possible to chain the invocations:
+//
+//	NewNode(transformNewNodeArgs(...))
+func transformNewNodeArgs(deprecationTracker *types.DeprecationTracker, nodeType types.NodeType, text string, args ...any) (*types.DeprecationTracker, types.NodeType, string, []any) {
```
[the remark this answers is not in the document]
pohly: I checked my changes in Kubernetes where I have unit tests. That's how I found that `slices.Backward` (still in the patch above, not anymore in the PR) influences stack unwinding and needs to be avoided.

core_dsl.go
pohly:
```
// If there is more than one registered transformer, then the most
// recently added ones get called first.
```

I suppose this can be resolved then.

onsi:
sounds like this is GTG.  I'll take another final pass look in a bit and pull it in.  If you have time and the inclination adding some documentation to the `docs/index.md` file would be great.  I'm thinking something after the [Advanced: Around Node](https://onsi.github.io/ginkgo/#advanced-around-node) section called "Advanced: Transforming Node Arguments" - but no worries if you'd rather I do it.

pohly:
I probably won't get to writing more docs this week.

onsi:
no worries i’ll take care of it.

onsi:
my plan is to write up the docs and cut a release today.

PR #1608, opened 2025-10-20
Bug Fix: Add GinkoTBWrapper.Attr() and GinkoTBWrapper.Output()

andrewradamis-vividseats:
Add `Attr()` and `Output()` to `GinkgoTBWrapper`. `GinkoTB().Output()` is currently causing `runtime error: invalid memory address or nil pointer dereference`

Related:
- https://github.com/onsi/ginkgo/commit/0a618d45


onsi:
thanks!

PR #1610, opened 2025-10-23
Fix paragraph position error

s-z-z:
It seems the paragraph about NodeArgsTransformer was inserted in the wrong place. After “Here are a few more patterns that AroundNode enables.”, the following two subparagraphs should describe AroundNode-related content, same as before. The advanced features of NodeArgsTransformer should not interrupt that section. I’ve moved it accordingly, though please double-check that this new position makes sense.


onsi:
welp!  thanks for catching and fixing that!

PR #1611, opened 2025-10-24
Fix syntax errors and typo

s-z-z:
Sorry for another small fix commit — not intentional. 
The docs are extensive (and very well-written, thanks), and after finishing the full review, I found a few typos as it is.


onsi:
❤️  thanks for the fixes!  Super appreciate it - the docs are almost book-length and getting additional eyes on 'em is *super* helpful!

PR #1624, opened 2025-12-05
ginkgo: fix data race

pohly:
Write at 0x00c000368528 by goroutine 25:
      os/exec.(*Cmd).Wait()
          os/exec/exec.go:926 +0x1a4
      github.com/onsi/ginkgo/v2/ginkgo/internal.runParallel.func2()
          github.com/onsi/ginkgo/v2@v2.27.2/ginkgo/internal/run.go:231 +0x37

    Previous read at 0x00c000368528 by goroutine 96714:
      github.com/onsi/ginkgo/v2/ginkgo/internal.runParallel.func1()
          github.com/onsi/ginkgo/v2@v2.27.2/ginkgo/internal/run.go:228 +0x3b
      github.com/onsi/ginkgo/v2/internal/parallel_support.(*ServerHandler).procIsAlive()
          github.com/onsi/ginkgo/v2@v2.27.2/internal/parallel_support/server_handler.go:133 +0x121

The problem is that cmd.ProcessState is not thread-safe. The solution is an atomic bool which gets checked instead and which is written once when the goroutine waiting for the command has determined that the command has exited.

### DIFF
--- a/ginkgo/internal/run.go
+++ b/ginkgo/internal/run.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+	"sync/atomic"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


onsi:
perfect, thanks!

PR #1625, opened 2025-12-05
ginkgo: report exit result in case of failure

pohly:
When a worker unexpectedly dies, it's hard to determine why. Even if output is available (--output-interceptor-mode=none), if it gets killed by the Linux OOM killer there's nothing in the process output about that.

Including a textual description of the exit status provides that information (here simulated with `killall -KILL e2e.test`):

    Output from proc 1:
        I1204 10:03:40.282670  196272 e2e.go:109] Starting e2e run "82689063-1787-40c3-be7d-c4677c556063" on Ginkgo node 1

    Exit result of proc 1:
      signal: killed
      
      
Real example:
- without this: https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/133844/pull-kubernetes-e2e-kind-alpha-beta-features-race/1996487725544378368/build-log.txt
- with this: https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/133844/pull-kubernetes-e2e-kind-alpha-beta-features-race/1996506959712358400/build-log.txt


onsi:
super helpful, thanks @pohly

PR #1638, opened 2026-01-15
Fix doclink for ginkgo run

sardobi:
The doclink for `ginkgo run` points to a nonexistent heading, `#running-tests`. Looks like it was renamed to `#running-specs`.

Let me know if you'd rather I opened an issue - thought this was a tiny enough change not to warrant one


onsi:
thank you!

PR #1650, opened 2026-04-04
Implement shell completion

koct9i:
This adds subcommand "completion" to the CLI which:
- generates completion script: ginkgo completion bash
- installs completion script: ginkgo completion --install bash
- generates commands completion: ginkgo completion --complete bash --
- generates flags completion: ginkgo completion --complete bash -- run --
- without suggests script fails back to default and completes file paths

For "bash" use hack similar to used in cobra, but implement it in go:
- single suggests are issued as is without description
- multiple suggests are issued in form "{suggest}{align}({description})" I.e. description is a part of suggest but shell will use only common prefix.

For "zsh" issue completion response in form "{suggest}:{description}".

For "fish" issue completion response in form "{suggest}\t{description}".

### DIFF
--- a/ginkgo/command/program.go
+++ b/ginkgo/command/program.go
@@ -1,9 +1,13 @@
 package command
 
 import (
+	"bufio"
 	"fmt"
 	"io"
+	"maps"
 	"os"
+	"path/filepath"
+	"slices"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


onsi:
cool, thanks!

PR #1658, opened 2026-05-03
Fix nested --github-output group for progress report nested inside timeline

koct9i:
Do not emit group when progress report is nested in timeline or failure.
Unfortunately github logging does not handle nested groups correctly.

Commit 606c1cb29be1b5e06c502e022559fb59e34d1629 passed "isVeryVerbose" as
to argument "emitGroup", while should pass to "emitGinkgoWriterOutput".
So, it break fix added in commit fdc65b13166ea27ef283a055289bfe3719318b17.


koct9i:
Funny enough my project still use 2.26.x, so I haven't seen that this feature is broken again since 2.27.x =)

onsi:
Thanks!

PR #1663, opened 2026-05-11
feature: devcontainer configuration for easier Ginkgo contribution

thediveo:
This adds an optional development container configuration to the Ginkgo repository:
- uses mcr.microsoft.com/devcontainers/base:ubuntu-24.04 as the base image.
- installs Go from upstream, as well as lazygit for all chores git.
- installs Ruby and then does a "bundle install" inside the docs directory, working around a problem with the missing bigdecimal dependency on newer Ruby installations.
- installs local pkgsite feature that services Ginkgo's godoc on local container port 6060.
- provides convenience commands:
  - `ginkgotest` runs the same Ginkgo-based full testing as the test workflow does.
  - `serve-docs` runs `bundle exec jekyll serve` inside the `docs/` directory, serving on local container port 4000.

To use with VSCode and the "dev containers" extension installed: command => "Dev Containers: Open Folder in Container...". This should automatically pick up the devcontainer configuration from `.devcontainer/devcontainer.json` and build the dev container according to spec (which will initially take a while); afterwards, the dev container is started and its configuration completed, wait for the bundler installation to finish.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,41 @@
+{
+    "name": "ginkgo",
+    "portsAttributes": {
+        "4000": {
+            "label": "Github pages documentation",
+            "protocol": "http"
+        },
+        "6060": {
+            "label": "Ginkgo package documentation",
+            "onAutoForward": "silent",
+            "protocol": "http"
+        }
+    },
+    "image": "mcr.microsoft.com/devcontainers/base:ubuntu-24.04",
+    "features": {
+        "ghcr.io/thediveo/devcontainer-features/local-pkgsite:0": {},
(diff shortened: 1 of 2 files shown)


onsi:
oh man, i'm an old school fart when it comes to fancy dev container tooling.  happy to pull this in if it makes other people's lives easier but might need help maintaining it and fielding issues if/when they come up.

thediveo:
Using devcontainers for (only) one and a half years I'm considering myself still kind of a late-comer, with the benefit of a more mature technology. For me, it simplifies working with different projects and the tools needed. I always struggle with Ginko's and Gomega's Ruby/Rails/Jeckyll setup and this gives me a quickly reproducable setup.

There are some startups with devcontainer tooling that doesn't work; devpod by loft labs doesn't work, they wanted to argue about what (transient) dependencies are.

PR #1664, opened 2026-05-13
Feature: GinkgoHelperGo

thediveo:
### Motivation

Based on discussion #1660 this implements a new `GinkgoHelperGo` (not to be confused with `GinkgoHelper`) that runs a helper function on a separate Go routine, waiting for it to either terminate successfully or fail. The typical usage is envisioned to be that the helper function runs a spec-supplied function on the new Go routine. A concrete example is running a test helper for remounting the Linux /sysfs inside mount and network namespaces different from those of the test binary. Because of the way mount namespaces work in Linux, a spec cannot do this "inline" but needs to carry out the /sysfs remount from a separate Go routine with its own locked OS-level thread that has been partly detached in terms of certain filesystem-related thread/process properties.

`GinkgoHelperGo` is Gomega-agnostic in that it passes a helper "Fail" instead of a helper "Gomega", yet Gomega is easily roped in.

I've simplified passing the outcome from the separate Go routine to the spec's Go routine, dropping the need for the discriminate "union" by simply using a `chan any` and passing either values of type `userPanicked` (string) or type `helperPanicked` (string), and an implicit `nil` by just closing the channel without sending anything as the success signal.  

### PR

- adds new `GinkgoHelperGo` -- I'm open to better naming ideas.
- `GinkgoHelperGo` is placed in a dedicated `helpergo_dsl.go` but I'll happily move it into `core_dsl.go` or other places, just tell me.
- updates both godoc as well as pages documentation regarding `GinkgoHelperGo`.
- adds integration tests.
- passes Ginkgo's complete test suite.


onsi:
hey thanks @thediveo - i'm away the next couple days but will try to take a look over the weekend!

onsi:
I like it!  Looks good to me :)

PR #1665, opened 2026-05-14
chore: refactor devcontainer for better maintenance

thediveo:
Sorry for the noise, this is hopefully making maintaining the devcontainer configuration much easier and clearer what's going on. This chore update now uses proper scripts instead of shoehorning non-trivial shell commands into JSON strings.

On top, I've made the first-run-message to be better readable and more comprehensible.

This is now basically the same devcontainer configuration as the Gomega PR, just with the appropriate name changes from Gomega to Ginkgo.


onsi:
all good!

Issue #1668, opened 2026-06-04
Feature Request: Emulate -fd style output


woodie:
## Emulate "format documentation" style output

Ginkgo's output can be a bit noisy for large test suites compared to RSpec's `-fd` output. The [`ginkgo-fd`](https://github.com/woodie/ginkgo-fd) post-processor is an attempt to achieves this by consuming Ginkgo's `--json-report` output.

Should I create a PR that adds the `-fd` flag directly into Ginko?

onsi:
sure thing!  go for it!  be sure to update the docs as well, thanks!

woodie:
We could use the same footer.
```
Ran 127 of 127 Specs in 0.078 seconds
SUCCESS! -- 127 Passed | 0 Failed | 0 Pending | 0 Skipped
```
vs this
```
Finished in 75ms
127 examples, 0 failures
```

onsi:
imo keep the same footer

woodie:
Will do. A direct clone seemed like a good starting point, but more (unnecessary) work to maintain.

woodie:
OK, the latest PR uses the same footer.

```
$ go run github.com/onsi/ginkgo/v2/ginkgo -fd  ./reporters/...
DefaultReporter
  Rendering SuiteWillBegin
    Default Behavior
    With Labels
    With long Labels
    With SemVerConstraints
    With ComponentSemVerConstraints
    When configured to randomize all specs
    when configured to run in parallel
    when succinct and in series
    when succinct and in parallel
    when succinct and with labels
    when succinct and with SemVerConstraints
    when succinct and with ComponentSemVerConstraints
•••
JunitReport
  default behavior
    generates a Junit report and writes it to disk
  when configured to omit all the omittables
    generates a Junit report and writes it to disk
  when configured to write the report inside a folder
    creates the folder and the report file

Ran 127 of 127 Specs in 0.078 seconds
SUCCESS! -- 127 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS

Ginkgo ran 1 suite in 2.338581833s
Test Suite Passed
```

woodie:
There was a bit to unwind from the first implementation, but change looks solid now.

woodie:
I'm also working on simplifying this change in [this PR](https://github.com/onsi/ginkgo/pull/1670).

Huaian666:
one of those rare projects where the more you use it the more you appreciate it. small things keep surprising me in a good way.

[...]

PR #1669, opened 2026-06-04
RSpec-style documentation output - short circuits

woodie:
This feature was proposed in issue [1668](https://github.com/onsi/ginkgo/issues/1668). From the updated docs...

You can opt into documentation format output with ginkgo -fd.  This emits a hierarchical tree of spec descriptions, with each spec's name colored green for passing, red for failing, yellow for pending, and cyan for skipped, along with a summary of failures at the end of the suite.  This mode will be familiar to Rspec users and has more concise output for very large test suites.

This PR may feel bolted on with several methods starting with a short cirtuit.
`if r.conf.FdOutput { return }`

PR #1670, opened 2026-06-05
RSpec-style documentation output

woodie:
The goal is better integration without bolting on. The issue is [here](https://github.com/onsi/ginkgo/issues/1668).

You can opt into documentation format output with ginkgo -fd.  This emits a hierarchical tree of spec descriptions, with each spec's name colored green for passing, red for failing, yellow for pending, and cyan for skipped, along with a summary of failures at the end of the suite.  This mode will be familiar to Rspec users and has more concise output for very large test suites.

This attempts to integrate with minimal changes and should work better with future refactoring.


onsi:
thanks @woodie - i think i see you’ve got two PRs going?  Let me know when the dust settles and you’d like me to take a look.

woodie:
> thanks @woodie - i think i see you’ve got two PRs going? Let me know when the dust settles and you’d like me to take a look.

Both PRs are almost  identical with slight changes to `reporters/default_reporter.go`:
 • [1669](https://github.com/onsi/ginkgo/pull/1669) may feel bolted on with several methods starting with a short cirtuit.`if r.conf.FdOutput { return }`
 • [1670](https://github.com/onsi/ginkgo/pull/1670) attempts to integrate with minimal changes and should work better with future refactoring.

I've renamed the PRs. I like the 2nd approach as only so many things can be bolted on without creating a mess.

woodie:
OK, I like this PR @onsi. I did post a couple questions in the Issue. The dust has settled for me.

woodie:
Hey @onsi, are we ready to merge this?

onsi:
hey sorry - i'd missed your earlier reply on the issue.

I think this will work great if the user simply runs `ginkgo  -fd`.  but if they run `ginkgo -p  -fd` or `ginkgo -randomize-all  -fd` it will generate interleaved output that won't make sense.  I think that's fine (changing the behavior is likely too complex - and doesn't match the use-case), but I think what we'd want is to have Ginkgo emit an error and refuse to run if you mix `-fd` with `-p`, `-numProcs=N` (with `N>1`), or `-randomize-all`.  Are you up for adding that guard?

Alternatively, if you've tried it with `-p` and `-randomize-all` and think the output is fine, lemme know!

woodie:
OK, I've got [this commit](https://github.com/onsi/ginkgo/pull/1670/changes/15b9f446664e196ed5a0a5a2aac6cca888079ad4), which is a bit noisy. Let me know if I need to tighten it up more.

PR #1673, opened 2026-06-11
Allow extensions/globals.Reset to support re-running RunSpecs

### LINKED ISSUE #1672: Support running multiple suites sequentially in one process via extensions/globals.Reset
## The need

Programs that drive Ginkgo programmatically sometimes want to run **several suites sequentially in a single process**: register one suite's specs, call `RunSpecs`, reset, register the next suite's specs, and call `RunSpecs` again. Two use cases we know of:

- **Merged test binaries in large monorepos.** When hundreds of packages each produce a near-identical statically-linked test binary, compiling one shared binary and selecting a suite at runtime saves enormous build time and disk. Each suite invocation then pays full process boot (runtime init, every package's `init()`, connection pools) — running suites back-to-back in one process amortises it.
- **Meta-testing of dynamically generated suites** — the use case `extensions/globals.Reset` was added for in #692.

## How close Ginkgo already is

Nearly everything supports this today:

- `extensions/globals.Reset()` swaps in a fresh `internal.Suite` and `Failer`.
- `Fail`/`Skip`/`AbortSuite` resolve `global.Failer` at call time, so registered fail handlers keep working across resets.
- `RunSpecs` constructs its per-run resources (reporter, output interceptor, parallel client, progress-signal handler) fresh each call, and restores the pre-run suite via `PushClone`/`PopClone`.

The one blocker is the `suiteDidRun` guard: it lives as an unexported package variable in `ginkgo`, where `Reset` (a different package) cannot clear it — so a second `RunSpecs` always exits with `GinkgoErrors.RerunningSuite()`, even after an explicit reset.

## Proposal

Move the flag into `internal/global` (as `global.SuiteDidRun`) and clear it in `InitializeGlobals()`. Calling `RunSpecs` twice **without** an intervening `Reset` still fails exactly as today — only an explicit `Reset` arms a fresh run, so the accidental-rerun protection is preserved.

The change is ~10 lines; we've prepared a PR with a test.

## Evidence it works

We run this in production CI at incident.io: a generated mega test binary absorbs ~690 ginkgo suites from our monorepo, and worker processes drain a suite queue with `Reset` + re-register + `RunSpecs` between suites. Per-suite reporting (including `--ginkgo.junit-report`) behaves identically to separate processes, and our CI test phase dropped from ~98s to ~47s.

lawrencejones:
Closes #1672.

## What

Programs that drive Ginkgo programmatically (merged test binaries for large monorepos, meta-testing of dynamically generated suites) want to run several suites sequentially in one process: register specs, call `RunSpecs`, `Reset`, register the next suite's specs, and call `RunSpecs` again.

Everything `RunSpecs` uses is already either reconstructed per run (reporter, output interceptor, parallel client, progress handler) or reset by `extensions/globals.Reset` (the suite and failer themselves). The one blocker is the `suiteDidRun` guard, which lives in package `ginkgo` where `Reset` cannot reach it, so a second `RunSpecs` always exits with `GinkgoErrors.RerunningSuite`.

This moves the flag into `internal/global` as `SuiteDidRun` and clears it in `InitializeGlobals`. Calling `RunSpecs` twice **without** an intervening `Reset` still fails exactly as before — only an explicit `Reset` arms a fresh run, preserving the accidental-rerun protection. Includes a test in `extensions/globals`.

## Why we want it

At incident.io we compile ~690 monorepo suites into one generated test binary (each per-suite binary was ~350MB of near-identical bytes). With this change, CI worker processes drain a suite queue — `Reset`, re-register, `RunSpecs` — paying process boot once per worker instead of once per suite. Per-suite reporting (`--ginkgo.junit-report` et al) behaves identically to separate processes, and our test phase dropped from ~98s to ~47s. We're currently running this in production CI from a fork pinned via `replace`, and would love to drop the fork.

### DIFF
--- a/core_dsl.go
+++ b/core_dsl.go
@@ -39,7 +39,6 @@ var flagSet types.GinkgoFlagSet
 var deprecationTracker = types.NewDeprecationTracker()
 var suiteConfig = types.NewDefaultSuiteConfig()
 var reporterConfig = types.NewDefaultReporterConfig()
-var suiteDidRun = false
 var outputInterceptor internal.OutputInterceptor
 var client parallel_support.Client
 
@@ -259,10 +258,10 @@ for more on how specs are parallelized in Ginkgo.
 You can also pass suite-level Label() decorators to RunSpecs.  The passed-in labels will apply to all specs in the suite.
 */
 func RunSpecs(t GinkgoTestingT, description string, args ...any) bool {
-	if suiteDidRun {
+	if global.SuiteDidRun {
 		exitIfErr(types.GinkgoErrors.RerunningSuite())
 	}
-	suiteDidRun = true
+	global.SuiteDidRun = true
 	err := global.PushClone()
 	if err != nil {
 		exitIfErr(err)
--- a/extensions/globals/globals_test.go
+++ b/extensions/globals/globals_test.go
@@ -24,3 +24,18 @@ func TestGlobals(t *testing.T) {
 		t.Error("got the same suite but expected it to be different!")
 	}
 }
+
+func TestResetAllowsRerunningSpecs(t *testing.T) {
+	// RunSpecs refuses to run twice against the same suite, tracked by
+	// global.SuiteDidRun. Reset must clear it alongside the suite itself, so a
+	// program can run several suites sequentially in one process: register specs,
+	// RunSpecs, Reset, register the next suite's specs, RunSpecs again.
+	global.InitializeGlobals()
+
+	global.SuiteDidRun = true
+	globals.Reset()
+
+	if global.SuiteDidRun {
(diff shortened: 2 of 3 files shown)


lawrencejones:
Shout if you need more info on this, I'm hoping it should be an easy one 🙏

onsi:
wow what a trivial but impactful change.  Do you mind updating (or asking a robot to update) the long-form docs as well?

I'll pull this in for now but a follow-on PR for the docs would be much appreciated.

lawrencejones:
Getting a docs PR prepared now, thank you for being so responsive!

lawrencejones:
Docs here: https://github.com/onsi/ginkgo/pull/1674

PR #1676, opened 2026-06-17
feat: add --sleep-on-failure to pause a failed spec before teardown

### LINKED ISSUE #1671: Feature request: --sleep-on-failure to pause a failed spec before teardown
**Problem**
When a spec fails against a real environment (cluster, DB, service), `AfterEach`/`DeferCleanup` tears down the very state I need to inspect before I can look at it. I want to pause a failed spec *before* cleanup runs so I can do forensics on the live system.

**Why existing features don't fit**
`--fail-fast` still runs teardown; `ContinueOnFailure` doesn't pause; a manual `time.Sleep` in `AfterEach` fires on every spec and isn't runtime-configurable.

**Proposal**
New flag `--sleep-on-failure=<duration>` (default 0 = off): on spec failure, wait that long *before* teardown, then continue. Skippable with Ctrl-C.

**Question**
Is there appetite for this? If so I'm happy to send a PR (flag + tests + docs).

qinqon:
Closes #1671

## What

Adds a `--sleep-on-failure=<duration>` flag (default `0` = off). The moment a spec fails, Ginkgo pauses for the configured duration **before** any teardown (`AfterEach`/`JustAfterEach`/`DeferCleanup`) runs, so you can inspect the live system before its state is torn down.

## Why

When a spec fails against a real environment (cluster, DB, service), teardown destroys the very state you need to debug the failure. The existing building block (`time.Sleep` in a top-level `JustAfterEach`) works but isn't runtime-configurable. This packages that need as a first-class, well-behaved flag.

## How

Per the review feedback, this hooks directly into the suite's existing failure-handling path rather than injecting any nodes:

- In `runNode`, when a node finishes with a failure, Ginkgo emits the failure as it already does and then — if `--sleep-on-failure` is set — pauses right there, before returning to the spec loop (and therefore before any teardown runs). It emits a progress report with the helper text telling you the suite is paused and what to do.
- Only failures in **setup and subject nodes** (`It`, `Before*`, `BeforeSuite`...) pause. Failures in teardown/cleanup or reporting nodes are skipped, since the system is already being torn down at that point. This is implemented by skipping `NodeTypesAllowedDuringCleanupInterrupt | NodeTypesAllowedDuringReportInterrupt`.
- The pause `select`s on a timer and `interruptHandler.Status().Channel`, so the pause is **interruptible**: pressing `^C` ends it early and the suite proceeds to run cleanup as usual rather than skipping it.

This addresses the three points raised in #1671:

1. **Clear messaging** — the failure is emitted, followed by a progress report telling you the suite is paused and to press `^C` to proceed to cleanup.
2. **`^C` proceeds to cleanup** — the pause ends on interrupt and teardown then runs (covered by tests).
3. **Parallel** — since Ginkgo's parallelism is multi-process, a single failing spec can't meaningfully freeze the whole system. Rather than pretend otherwise, combining `--sleep-on-failure` with `-p`/`--procs` is rejected with a configuration error. This is intended as a serial debugging aid; run the failing spec serially (e.g. with `--focus`).

Also updates godoc and `docs/index.md` (under the interrupt/grace-period material).

## Test plan

- [x] `go vet ./...` passes
- [x] New `internal/internal_integration` tests: pause happens at the point of failure before teardown, setup-node (`BeforeEach`) failures pause, teardown-node failures do **not** pause, `^C` ends the pause and cleanup still runs, passing specs are never paused, and the flag is a no-op when `0`
- [x] New `integration` (CLI subprocess) tests: serial run waits ~the configured duration with ordering `failing body → pause notice → teardown`; parallel run exits with the serial-only error and does not hang
- [x] Full suite passes (`go test ./...`)

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -3305,6 +3305,24 @@ One final, somewhat complex, note on timeouts and the Grace Period.  As mentione
 - If the remaining node is interruptible and **does not** have a `NodeTimeout`, Ginkgo uses the Grace Period to set a deadline for the node.  If the deadline expires, then a second Grace Period applies before Ginkgo leaks the node and moves on.
 - If the remaining node is **not** interruptible, Ginkgo will give the node a single Grace Period to complete and exit.  In this case, since it cannot be interrupted, Ginkgo will simply leak the node after one Grace Period.
 
+#### Pausing a Failed Spec for Debugging: The --sleep-on-failure flag
+
+When a spec fails against a real environment (a cluster, database, or service) the spec's teardown - its `AfterEach`, `JustAfterEach`, and `DeferCleanup` nodes - typically tears down the very state you need to inspect to understand the failure.  By the time you can attach a debugger or run `kubectl`/`psql`, the evidence is gone.
+
+The `--sleep-on-failure=<DURATION>` flag pauses a spec **after it fails but before its teardown runs**, leaving the system live so you can perform forensics:
+
+```bash
+ginkgo --sleep-on-failure=30m ./...
+```
+
+With this flag set, the instant a spec fails Ginkgo emits the failure and then pauses for the configured duration before continuing.  The pause happens at the point of failure - before any teardown (`AfterEach`/`JustAfterEach`/`DeferCleanup`) runs - so the failing spec's resources remain in place for you to inspect.  Ginkgo emits a progress report announcing the pause so you know the suite is waiting and what to do next.  Failures that occur in setup and subject nodes (`It`, `BeforeEach`, `BeforeAll`, ...) trigger the pause; failures inside teardown nodes do not, since the system is already being torn down at that point.
+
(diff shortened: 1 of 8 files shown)


onsi:
hey there, can you ask opus to try again with some different guidance?  I'd prefer a different implementation that hooks into the suite's existing codepaths for handling failures.  If a failure is identified the suite should - at that moment - pause.  It should emit the full failure message, and then emit the helper text telling you to interact asking you to interact with the code.

Injecting a `JustAfterEach` is making some assumptions about test-ordering that may not hold.  It's also just adding more complexity than need be, I think the aforementioned approach would be better.

Thanks!

qinqon:
> hey there, can you ask opus to try again with some different guidance? I'd prefer a different implementation that hooks into the suite's existing codepaths for handling failures. If a failure is identified the suite should - at that moment - pause. It should emit the full failure message, and then emit the helper text telling you to interact asking you to interact with the code.
>

Sure, let me refactor this with that.
 
> Injecting a `JustAfterEach` is making some assumptions about test-ordering that may not hold. It's also just adding more complexity than need be, I think the aforementioned approach would be better.
> 

At the end I was kind of productifying how I hack my way into this.

> Thanks!

Happy to contribute, I have being using ginkgo for a long long time now.

onsi:
LGTM - thanks!

onsi:
i'll cut a release later today/tomorrow

PR #1678, opened 2026-07-01
Properly Punctuate Possessive Pronoun "Its"

cunnie:
The possessive form of the pronoun "it" ("its") never takes an apostrophe, and yet the documentation had three cases with the heretical punctuation. With this commit we restore orthodoxy to the docs.

[This is a minor commit so I didn't clutter your work by opening an issue even though CONTRIBUTING suggests I do]


onsi:
😂

thanks!

PR #1681, opened 2026-08-10
Defer AfterAll until repeated spec completes

### LINKED ISSUE #1680: `AfterAll` fired earlier and multiple times when `It` with a `MustPassRepeatedly` decorator is the last element
## When I

- Use `AfterAll`

## I expect

- To be fired **after all** tests, at the end.
- To be fired **once**.

## Instead

- If the last `It` element has `MustPassRepeatedly`, it's fired after each execution.
- `AfterAll` is no longer a clean-up step, but it becomes an `AfterEach` for that `It` element.
- Order is now extremely important, and the behavior is misleading.

## Notes

Context:

- We're using Ginkgo for a testsuite.
- In the testsuite there are stability test, where a client is opened in `BeforeAll` and closed in `AfterAll`.
- We put one of those stability test at the end.
- `AfterAll` now closes the client after the first execution of the stability test (`It`) even though it needs to be repeated (marked as `MustPassRepeatedly`)

### Repro

> [!NOTE]
> **AI Disclosure**: the code to reproduce the issue has been generated by an AI and reviewed by me.

Code to reproduce the issue: https://github.com/ilovelinux/ginkgo/tree/issue-repro-afterall-mustpassrepeatedly

Boulea7:
Fixes #1680.

`AfterAll` is a run-once teardown for an `Ordered` container. When the final spec used `MustPassRepeatedly`, each intermediate successful attempt was treated as the last opportunity to run teardown. This caused `AfterAll` to run before the next attempt and potentially more than once.

This change bases the run-once after-node decision on whether the active repeat or retry policy will schedule another attempt. A passing `MustPassRepeatedly` attempt defers teardown while more required passes remain. For the final spec, teardown runs after all required passes, or after an early failure when repetition stops. `FlakeAttempts` keeps its existing inverse behavior: failed attempts defer teardown while they can be retried, and a passing attempt completes it.

The integration coverage records exact attempt events and node order for:

- nested non-final and final repeated specs
- a repeated spec that fails before exhausting its attempts
- `BeforeAll`/`AfterAll` and per-attempt `BeforeEach`/`AfterEach`
- parallel and `Serial` ordered containers

Validation performed locally with Go 1.26.5:

- focused ordered and parallel integration tests
- the focused race-enabled integration tests in 20 consecutive runs
- all `internal/...` packages other than the full `internal_integration` package
- `go vet ./...`
- gofmt, diff checks, and incremental golangci-lint with no new issues

The full `internal_integration` package reports 369 of 370 specs passing. The remaining unchanged timeout test compares a `CodeLocation` whose `FullStackTrace` is populated while the assertion expects it to be empty. The same focused failure reproduces at the unmodified base commit `09bbf5f` with Go 1.26.5.


onsi:
code lgtm thanks!  Will pull this in once CI goes green.

Boulea7:
Thanks for the review! I'm checking the oldstable integration failure now and will update once I understand what's happening.

onsi:
cool, thanks.    if it's an unrelated flake don't worry about it.

ilovelinux (approved):
I double-checked the implemented tests empirically:

```console
$ git remote add -f Boulea7 https://github.com/Boulea7/ginkgo
$ git switch fix/ordered-afterall-must-pass-repeatedly
$ git restore --source master internal/group.go # Reverse the bugfix
$ ginkgo -r -p # seed: 1786398309
[...]
Summarizing 4 Failures:
  [FAIL] Running tests in parallel [It] distributes tests across the parallel procs and runs them
  /home/user/projects/github/onsi/ginkgo/internal/internal_integration/parallel_test.go:159
  [FAIL] Ordered Containers [It] when the final spec must pass repeatedly, it runs BeforeAll and AfterAll just once
  /home/user/projects/github/onsi/ginkgo/internal/internal_integration/ordered_test.go:61
  [FAIL] Ordered Containers [It] when a non-final spec must pass repeatedly, it runs BeforeAll and AfterAll just once
  /home/user/projects/github/onsi/ginkgo/internal/internal_integration/ordered_test.go:61
  [FAIL] Ordered Containers [It] when the final spec fails while it must pass repeatedly, it runs AfterAll after the failure
  /home/user/projects/github/onsi/ginkgo/internal/internal_integration/ordered_test.go:61

Ran 370 of 370 Specs in 2.579 seconds
FAIL! -- 366 Passed | 4 Failed | 0 Pending | 0 Skipped
[...]
```

They fail as expected! 😄 

LGTM!

onsi:
just shipped 2.32.1 with this in it.  thanks!

Boulea7:
That's great to hear—thank you for merging it and shipping the fix in v2.32.1!

PR #1682, opened 2026-09-09
junit: support ReportEntry

pohly:
Some JUnit tools support a "properties" element at the "testcase" level. This can be used to stored the ReportEntries attached to a spec. The string representation was chosen because the JUnit format seems to imply that this meant to be human-readable.

### DIFF
--- a/reporters/junit_report.go
+++ b/reporters/junit_report.go
@@ -11,6 +11,7 @@ The schema used for the generated JUnit xml file was adapted from https://llg.cu
 package reporters
 
 import (
+	"encoding/json"
 	"encoding/xml"
 	"fmt"
 	"maps"
@@ -136,6 +137,8 @@ type JUnitTestCase struct {
 	SystemOut string `xml:"system-out,omitempty"`
 	//SystemOut maps onto any captured GinkgoWriter output - maps onto SpecReport.CapturedGinkgoWriterOutput
 	SystemErr string `xml:"system-err,omitempty"`
+	//Properties captures any ReportEntries attached to the spec via AddReportEntry, as key/value pairs with the JSON dump of the value as value.
+	Properties *JUnitProperties `xml:"properties,omitempty"`
 }
 
 type JUnitSkipped struct {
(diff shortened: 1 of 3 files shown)


reporters/junit_report.go
```diff
@@ -241,6 +243,16 @@ func GenerateJUnitReportWithConfig(report types.Report, dst string, config Junit
 		if !config.OmitCapturedStdOutErr {
 			test.SystemOut = systemOutForUnstructuredReporters(spec)
 		}
+		if len(spec.ReportEntries) > 0 {
+			properties := JUnitProperties{}
+			for _, entry := range spec.ReportEntries {
+				properties.Properties = append(properties.Properties, JUnitProperty{
+					Name:  entry.Name,
+					Value: entry.StringRepresentation(),
```
pohly: There's one potential issue: this will prefer the colored string if available, which is less suitable here than plain string. Is this worth addressing? Would imply changing StringRepresentation.

Or use a JSON dump as string after all?

onsi: My instinct is to go with the JSON dump as the string as that enables structured objects to flow into the reports more naturally and it would align with what the Ginkgo JSON report does.  If you want a non-colored string though then I'dd add a new `entry.NoColorStringRepresentation()`

onsi:
hey @pohly this LGTM - i replied to your comment.  feel free to take it in whatever direction you'd like. just let me know when you're happy with it and i'll pull it in and cut a release.

reporters/junit_report.go
pohly: Okay, let's go with JSON then. I need to check how to get to that, though - if I remember correctly, it needs a new method.

pohly: Done: https://github.com/onsi/ginkgo/compare/23fa08bba6c06aff32df6042b351853b5a40403c..3dad23be5a5e269580b4604c456493c5d7e26176

`json.Marshal(entry.GetRawValue())` works without any new helper method.

onsi: yep that's right.  it mirrors `ReportEntryValue.MarshalJSON()`

